PR #6, opened 2023-07-01
Save and display timeToComplete on model outputs

arcticfly:
Before:
Screenshot 2023-06-30 at 8 41 03 PM

After:
Screenshot 2023-06-30 at 8 40 42 PM

### DIFF
--- a/prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
+++ b/prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
@@ -1,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `timeToComplete` to the `ModelOutput` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "ModelOutput" ADD COLUMN     "timeToComplete" INTEGER NOT NULL;
--- a/prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
+++ b/prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
@@ -1,8 +1,0 @@
-/*
-  Warnings:
-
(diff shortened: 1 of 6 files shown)


prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `timeToComplete` to the `ModelOutput` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "ModelOutput" ADD COLUMN     "timeToComplete" INTEGER NOT NULL;
```
corbt: Pretty sure this migration will fail in prod without a default value?

src/server/utils/getChatCompletion.ts
```diff
@@ -25,10 +27,12 @@ export async function getChatCompletion(
     output: Prisma.JsonNull,
     errorMessage: null,
     statusCode: response.status,
+    timeToComplete: 0
   };
 
   try {
     resp.output = await response.json();
+    resp.timeToComplete = Date.now() - start;
```
corbt, line 34: nit: should probably be before awaiting the json because that could take an unknown amount of time if the thread loses priority.

prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
corbt: `timeToComplete` should probably be optional anyway since if a request fails it won't be set.

src/server/utils/getChatCompletion.ts
arcticfly: Sounds good

prisma/migrations/20230701030641_add_time_to_complete_to_model_output/migration.sql
arcticfly: Good point, I added a default of 0

PR #7, opened 2023-07-01
Add AutoResizeTextArea

arcticfly:
Before, long variant header titles were cut off.

Before:
Screenshot 2023-07-01 at 7 47 08 AM

After:
Screenshot 2023-07-01 at 7 47 30 AM


corbt (approved):
Good change!

PR #9, opened 2023-07-04
Add streaming for normal text

arcticfly:
This PR adds support for streaming normal text responses to the UI by creating a new websocket server, emitting events from the NextJS server to the websocket server, and reading those events from the client's websocket connection. A new websocket is created for each prompt variant * scenario id combination, which we may want to revisit later, but for now it shouldn't be a problem at our small scale.


corbt (approved):
Looks great! I'll merge and work on getting it to deploy.

PR #15, opened 2023-07-07
Retry requests that receive 429

arcticfly:
## Changes
* Retry requests automatically on 429 errors and manually on any error
* Change OutputCell query to upsert
* Move OutputCell to its own folder, define child components in their own files
* Create sharedStrings file, define 429 error string for both client and server

Before:
Screenshot 2023-07-06 at 9 23 12 PM

After:
Screenshot 2023-07-06 at 9 38 17 PM


corbt (commented):
I was imagining was that the retry logic would live in OutputCell. That way we can message the user that a call got rate limited, and we're retrying in X seconds

I think this one also doesn't have a way to let you retry again once it fails 3 times?

arcticfly:
> I was imagining was that the retry logic would live in OutputCell. That way we can message the user that a call got rate limited, and we're retrying in X seconds
> 
> I think this one also doesn't have a way to let you retry again once it fails 3 times?

Good idea.

src/components/OutputsTable/OutputCell/ErrorHandler.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {msToWait > 0 && (
+        <Text color="red.600" fontSize="sm">
+          Retrying in {Math.ceil(msToWait / 1000)} seconds...
+        </Text>
+      )}
+    </VStack>
+  );
+};
+
+function breakLongWords(str: string | null): string {
```
corbt, line 84: I assume this is so the text doesn't escape the container? using CSS for this like `word-break: break-all` handle that case more simply? https://css-tricks.com/almanac/properties/w/word-break/#aa-values

corbt (approved):

src/components/OutputsTable/OutputCell/ErrorHandler.tsx
arcticfly: break-all can break normal words as well, which we don't want

PR #16, opened 2023-07-07
Break error words without breakLongWord

arcticfly:
`word-break: break-word` does what we want, so we don't need `breakLongWord`.


corbt:
Yay deleting code!

PR #18, opened 2023-07-07
Change website layout

arcticfly:
This is the first part of the work to make the website look better on mobile (and everywhere else as well). The experiments page allows us to remove the experiment links from the left side navbar, which will allow us to contract the bar without too much trouble.

Navbar before:
Screenshot 2023-07-07 at 1 38 28 PM

Navbar after:
Screenshot 2023-07-07 at 1 38 38 PM

Experiments page:
Screenshot 2023-07-07 at 1 38 54 PM


src/server/api/routers/experiments.router.ts
```diff
@@ -4,11 +4,37 @@ import { prisma } from "~/server/db";
...
 export const experimentsRouter = createTRPCRouter({
   list: publicProcedure.query(async () => {
-    return await prisma.experiment.findMany({
+    const experiments = await prisma.experiment.findMany({
       orderBy: {
         sortIndex: "asc",
       },
     });
+
+    const experimentsWithCounts = await Promise.all(
```
corbt, line 14: This is ugly (vs doing an aggregate query) but we're only running this from the experiments index page now, not every page? Should be fine in that case.

src/utils/dayjs.ts
```diff
@@ -0,0 +1,16 @@
+import dayjs from "dayjs";
+import pluralize from "pluralize";
+
+export const formatTimePast = (date: Date) => {
```
corbt, line 8: I think you're looking for https://day.js.org/docs/en/durations/humanize :)

corbt (approved):
Looks much nicer!

src/server/api/routers/experiments.router.ts
arcticfly: I believe that is the case. I'll add a TODO.

src/utils/dayjs.ts
arcticfly: Yes, that helps with formatting, thanks!

PR #24, opened 2023-07-08
Record experiment updated in more places

arcticfly:
Previously we were only updating an experiment's `updatedAt` time when a new variant or scenario was created. We're now doing it every time a write command occurs against a variant or scenario.

## Changes
* Create new `recordExperimentUpdated` utility function
* Record updates for all write operations against variants and scenarios

### DIFF
--- a/src/server/api/routers/promptVariants.router.ts
+++ b/src/server/api/routers/promptVariants.router.ts
@@ -99,7 +99,7 @@ export const promptVariantsRouter = createTRPCRouter({
           })
         )._max?.sortIndex ?? 0;
 
-      const newVariant = await prisma.promptVariant.create({
+      const createNewVariantAction = prisma.promptVariant.create({
         data: {
           experimentId: input.experimentId,
           label: `Prompt Variant ${largestSortIndex + 2}`,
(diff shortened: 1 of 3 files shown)


corbt (approved):
It's a small thing, but I think I'd take the extra step of either (1) putting the two calls inside a $transaction, or (2) updating the experiment's `updatedAt` directly in the same call where the child update is made (see an example [here](https://stackoverflow.com/a/70806772)).

Since these changes block the UI adding more calls in serial could actually affect the UX if we do it too freely.

arcticfly:
> It's a small thing, but I think I'd take the extra step of either (1) putting the two calls inside a $transaction, or (2) updating the experiment's `updatedAt` directly in the same call where the child update is made (see an example [here](https://stackoverflow.com/a/70806772)).
> 
> Since these changes block the UI adding more calls in serial could actually affect the UX if we do it too freely.

Sure, let me give that a look.

arcticfly:
Okay, the experiment's `updatedAt` value should now be updated in the same transaction as whatever other write operation was previously occurring last in the write flow.

PR #38, opened 2023-07-14
Use javascript functions for prompt completions instead of templated json

### LINKED ISSUE #37: Proposal: refactor prompt templates to be functions instead of templated objects
I'm proposing a fairly major change to our concept of prompt templates. Instead of them being JSON files with string interpolation, they'll be small Javascript functions that take in a scenario and output a prompt.

Currently, prompt templates are defined as JSON documents with handlebars-style templating. For simple prompts that are constructed using only variable interpolation, they work well. However, in my experience many prompts benefit from more complex construction techniques. For example, it often makes sense to provide several example generations to the model as assistant prompts as part of the context. With our templating approach this is difficult to achieve.

I wanted to get a sense for how serious this limitation is, so I searched for open-source projects using the OpenAI API on Github. From the first 100 search results, I found [16 that seem to be real projects](https://corbt.notion.site/8917d4fbea8048648de4bb08b9e4fe04?v=c965cca3db884b15988419f450c459cb&pvs=4). 7/16 construct their prompts using basic string interpolation, while 9/16 use some more complicated logic (most commonly including a variable number of examples).

Switching prompt templates to be functions that produce objects instead of templated objects should make OpenPipe useful for a wider variety of prompts.

corbt:
Closes https://github.com/OpenPipe/OpenPipe/issues/37

This is a major change to the PromptVariant model. Instead of defining them as template objects with handlebars-style interpolations, it defines them as a Javascript function that takes a scenario and outputs a prompt template. This gives us a lot more flexibility in prompt construction at the cost of some complexity in both the execution environment and mental model.

Outstanding TODOs:
- [ ] Strip Typescript types from the user-provided functions before execution
- [ ] Ensure that the code executes and creates a valid object when passed a null scenario before saving the config
- [ ]  Get the model and save it to the prompt variant
- [ ] Get cost estimates working again


src/state/variantEditor.slice.ts
```diff
@@ -0,0 +1,87 @@
+import { type RouterOutputs } from "~/utils/api";
+import { type SliceCreator } from "./store";
+import loader from "@monaco-editor/loader";
+import openAITypes from "~/codegen/openai.types.ts.txt";
+
+export type VariantEditorSlice = {
```
arcticfly: Nit: I would maybe name this something like "VariantEditorModelSlice", since it doesn't contain information that we would store in the variant editor itself, such as the prompt-construction function.

arcticfly (approved):
A lot of good work here. I think you're right that we should merge ASAP with buggy code to minimize conflicts.

src/state/variantEditor.slice.ts
corbt: renamed to `sharedVariantEditor` to hopefully make it more clear this is general init code not a specific editor instantiation!

Issue #33, opened 2023-07-10
Move LLM API call triggers from UI to server


arcticfly:
We currently trigger calls to LLM API's in `OutputCell` in the UI. We should move these triggers to the server so that we don't have to render every cell in order to run our experiment against all scenarios and prompt variants.

Blocks: https://github.com/OpenPipe/OpenPipe/issues/34

## Architectural Changes:

At a high level, we need to do the following:
* Add Bull jobs to make requests to LLMs and retry if they fail
* Create a streaming channel id on the server instead of the frontend as we currently do so that requests can start with streaming enabled, even if the user is not currently viewing the model output in question
* Poll the server from in `OutputCell` on the frontend

#### Existing flow
1. New scenario or variant on the experiment generates new scenario-variant combination
2. Website invalidates experiment data in cache and requests updated data, which includes new cells
3. Website renders `OutputCell`s for each scenario-variant combination
4. Each `OutputCell` makes a request to the server for its corresponding `ModelOutput` in the database. We return here if either of the conditions are met:
   1. If the scenario-variant key exists in the database, it's immediately returned
   2. If the `inputHash` of the scenario-variant exists in the database (meaning that another scenario-variant has an identical templatized prompt), a new `ModelOutput` for this scenario-variant is generated with the same output as the existing `ModelOutput`'s output values
5. Otherwise, we call `getChatCompletion` and pass the `channel` if the frontend supplied one for streaming purposes
6. We forward the templated prompt to the correct LLM, and stream completions to the websocket server if they come in. In the event of rate limiting, the server returns a rate limit error, and the frontend repeats step 4 up to 3 times
7. When the model output is generated, we create a new `ModelOutput` and return it. The website invalidates prompt stats, so that cost and evaluation metrics can be updated

Our new process will generate output and update prompt stats without relying on all `OutputCell`s being rendered.

#### New flow
1. New scenario or variant on the experiment results in new scenario-variant combination
2. For each new scenario-variant combination, one of the following occurs:
    1. If the scenario-variant's `inputHash` exists in the database, we generate a new `ModelOutput` with identical output values to the existing `ModelOutput`
    2. If the `inputHash` does not exist, we create a new `ModelOutput` row in our database marked as `PENDING` with empty output values and queue a Bull job to handle requests to the LLM.
[...]

PR #39, opened 2023-07-14
Trigger llm output retrieval on server

arcticfly:
Implements https://github.com/OpenPipe/OpenPipe/issues/33

This PR includes changes to both the frontend and backend. On the frontend, cells switch from triggering LLM completions via mutations to pure polling. On the backend, LLM completions are triggered when new experiments, prompt variants, or scenarios are created or updated. We use graphile workers to make the requests to LLM providers asynchronously, so our prompt and scenario creation requests can return early.

After this PR is merged, devs should run `npx prisma migrate dev && tsx src/server/scripts/migrateScenarioVariantOutputData.ts` to migrate their data.

## Changes
* Rename `ModelOutput` table to `ScenarioVariantCell`, make output variables optional
* Create new `ModelOutput` table with only output variables
* Add `CellRetrievalStatus` enum
* Add `migrateScenarioVariantOutputData.ts` script to migrate existing data to new schema
* Change `OutputCell` to only poll, add hard refetch arrow in top right
* Add graphile workers to carry out LLM calls
* Increase `MAX_AUTO_RETRIES` from 3 to 10

Future work:
* Fix seeding
* Reschedule jobs for the future
* Properly invalidate prompt statistics


prisma/schema.prisma
```diff
@@ -75,17 +75,28 @@ model TemplateVariable {
...
+model ScenarioVariantCell {
     id String @id @default(uuid()) @db.Uuid
 
-    inputHash      String
-    output         Json
-    statusCode     Int
-    errorMessage   String?
-    timeToComplete Int     @default(0)
+    inputHash        String
+    output           Json?
```
corbt: I'm a little confused on the relationship between ScenarioVariantCell and ModelOutput. Why are `output`, `completionTokens`, `statusCode`, `timeToComplete`, `testScenarioId`, `promptVariantId` defined on both of them? Would it be possible to keep `promptVariantId` and `testScenarioId` on `ScenarioVariantCell`, and the rest on `ModelOutput`? That might make the queries a little more complex but shouldn't be too bad, and having a single source of truth for each of those fields should make the mental model of where to look for each piece of information clearer.

src/server/tasks/queryLLM.task.ts
```diff
@@ -0,0 +1,142 @@
...
+    where: { id: scenarioVariantCellId },
+    include: { modelOutput: true },
+  });
+  if (!cell) {
+    throw new Error("Cell not found");
+  }
+
+  // If cell is not pending, then some other job is already processing it
+  if (cell.retrievalStatus !== "PENDING") {
+    throw new Error("Cell is not pending");
```
corbt: Is this really an error condition? Shouldn't we just return in this case?

In any case, a job throwing an error is typically a signal to the runner that it should retry the job later. Is that really what we want in this case?

prisma/schema.prisma
```diff
@@ -100,6 +111,26 @@ model ModelOutput {
...
+    testScenarioId   String       @db.Uuid
+
+    createdAt DateTime @default(now())
+    updatedAt DateTime @updatedAt
+
+    scenarioVariantCellId String        @db.Uuid
+    scenarioVariantCell   ScenarioVariantCell @relation(fields: [scenarioVariantCellId], references: [id])
+
+    @@unique([scenarioVariantCellId])
+    @@index([scenarioVariantCellId])
```
corbt: Nit: you don't need this, `@@unique` already implies `@@index`

prisma/schema.prisma
```diff
@@ -75,17 +75,28 @@ model TemplateVariable {
...
+
+model ScenarioVariantCell {
     id String @id @default(uuid()) @db.Uuid
 
-    inputHash      String
-    output         Json
-    statusCode     Int
-    errorMessage   String?
-    timeToComplete Int     @default(0)
+    inputHash        String
```
corbt: I think the code will be conceptually cleaner if you move `inputHash` from `ScenarioVariantCell` to `ModelOutput`.

src/server/api/routers/scenarioVariantCells.router.ts
```diff
@@ -0,0 +1,71 @@
...
+        },
+      });
+
+      if (!cell) {
+        await generateNewCell(input.variantId, input.scenarioId);
+        return true;
+      }
+
+      if (cell.modelOutput) {
+        await prisma.modelOutput.delete({
```
corbt, line 55: Longer term we probably want to keep these around and expose them in the UI as previous generations. Would it be a problem to keep these and just show the latest one for now?

corbt (commented):
I'm somewhat confused about the data model here. I like the idea of separating a ScenarioVariant from a ModelOutput, because that'll let us run the same scenario x prompt multiple times potentially and show it to the user, and gives us a convenient place to attach eval results when we need to.

But the actual relationship between ScenarioVariantCell and ModelOutput in this code isn't totally clear to me and I think it would be cleaner to have less duplication between the fields on those models.

src/server/tasks/queryLLM.task.ts
```diff
@@ -0,0 +1,142 @@
...
+    const delay = calculateDelay(i);
+    await prisma.scenarioVariantCell.update({
+      where: { id: cellId },
+      data: {
+        errorMessage: "Rate limit exceeded",
+        statusCode: 429,
+        retryTime: new Date(Date.now() + delay),
+      },
+    });
+    await sleep(delay);
```
corbt, line 46: It probably doesn't matter for now but we should probably switch to requeuing the job in the future with the [`runAt` parameter](https://github.com/graphile/worker#addjob) set instead of sleeping within a job. That way if there are other jobs waiting to run they can run while we're waiting on this one.

prisma/schema.prisma
arcticfly: I agree explanation here is in order. I'll add comments to schema.prisma that explain why I put fields where I did, and we can go from there.

src/server/tasks/queryLLM.task.ts
arcticfly: No, we don't want to retry the job. I'll return instead.

prisma/schema.prisma
arcticfly: Removed

prisma/schema.prisma
arcticfly: Sure, that works

src/server/api/routers/scenarioVariantCells.router.ts
arcticfly: I think that's a valid thought, but probably something I want to do in a future PR (this one's already pretty complex).

arcticfly: I'll add a TODO

src/server/tasks/queryLLM.task.ts
arcticfly: Okay, I'll add a TODO

prisma/schema.prisma
arcticfly: I added comments to a few fields on the ScenarioVariantCell that will be deleted once the data migration is complete.

corbt:
Ok this is good enough, let's get it in and iterate from there.

PR #57, opened 2023-07-18
Allow user to duplicate prompt

arcticfly:
This change is the first step to allow the user to choose new models for their prompts.

## Changes
* Add drop-down menu for duplicating/hiding a prompt
* Change promptVariants `create` route to optionally duplicate an optionally provided variant

Before:
Screenshot 2023-07-18 at 1 43 05 PM

After:
Screenshot 2023-07-18 at 1 42 55 PM


corbt:
Ok yeah putting all three actions in one menu probably makes sense. Let's try this out and see if it feels annoying.

PR #63, opened 2023-07-19
Refine prompt

arcticfly:
Add a prompt variant dropdown option for refining prompt based on instructions provided by the user.

## Changes
* Add `getRefinedPromptFn` on promptVariants router
* Add RefinePromptModal

Modal:
Screenshot 2023-07-19 at 3 16 47 PM


src/components/OutputsTable/VariantEditor.tsx
```diff
@@ -22,7 +22,13 @@ export default function VariantEditor(props: { variant: PromptVariant }) {
     setIsChanged(currentFn.length > 0 && currentFn !== lastSavedFn);
   }, [lastSavedFn]);
 
-  useEffect(checkForChanges, [checkForChanges, lastSavedFn]);
```
arcticfly, line 25: @corbt Was this important? I can see it being beneficial if multiple people are working on the same prompt (though even then I think I'd prefer to update the user's editor to match the server so they don't accidentally overwrite their collaborator's saved work)

corbt: This is to show/hide the "save" button depending on whether your editor matches the last saved state. If you remove it I don't think the "save" button will show/hide reliably?

corbt (commented):
UX wise might it be better to just update the editor buffer with the refined prompt function, and then let the user tweak it and save when ready? That will save them a round of generations if they want to accept the refined prompt with some suggestions.

Don't feel strongly tho.

PR #89, opened 2023-07-24
Allow experiment forking

arcticfly:
This change let the user automatically fork another user's (or their own) experiment. This entails copying the experiment and all its variants, scenarios, cells, outputs, evaluations, output evaluations, and template variables.

## Changes
* Add `fork` route on experiments router
* Change plural prisma relations to camelCase
* Move `DeleteButton` to its own component
* Move `DeleteButton` to experiment settings drawer
* Add `HeaderButtons` with menu

View without access:
Screenshot 2023-07-24 at 6 02 09 PM

View with access:
Screenshot 2023-07-24 at 6 01 35 PM

Experiments loading screen:
Screenshot 2023-07-24 at 4 21 32 PM

Experiment Settings Drawer:
Screenshot 2023-07-24 at 6 03 13 PM


src/server/api/routers/experiments.router.ts
```diff
@@ -77,6 +79,196 @@ export const experimentsRouter = createTRPCRouter({
     };
   }),
 
+  fork: protectedProcedure.input(z.object({ id: z.string() })).mutation(async ({ input, ctx }) => {
```
arcticfly, line 82: @corbt relation copying logic if you're interested

src/components/experiments/HeaderButtons/useOnForkButtonPressed.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const user = useSession().data;
+  const experiment = useExperiment();
+
+  const forkMutation = api.experiments.fork.useMutation();
+
+  const [onFork, isForking] = useHandledAsyncCallback(async () => {
+    if (!experiment.data?.id) return;
+    const forkedExperiment = await forkMutation.mutateAsync({ id: experiment.data.id });
+    await utils.experiments.list.invalidate();
```
corbt: Do we need this? Adds latency and I actually think refreshing the experiment list right before going to the new one is more confusing than just going directly.

src/components/experiments/HeaderButtons/useOnForkButtonPressed.tsx
```diff
@@ -0,0 +1,64 @@
...
+      const currentQueries = queryString.parse(window.location.search);
+
+      // Set the fork parameter to true
+      const newQueries = queryString.stringify({
+        ...currentQueries,
+        fork: "true",
+      });
+
+      signIn("github", {
+        callbackUrl: `${window.location.origin}${window.location.pathname}?${newQueries}`,
```
corbt: just verifying you've confirmed this works and redirects correctly back to the page, right?

src/pages/experiments/[id].tsx
```diff
@@ -118,6 +45,10 @@ export default function Experiment() {
     }
   }, [updateMutation, experiment.data?.id, experiment.data?.label, label]);
 
+  useEffect(() => {
```
Comment (corbt) [inline src/pages/experiments/[id].tsx]: Was this moved just because you found the same bug before I merged https://github.com/OpenPipe/OpenPipe/pull/88?

src/server/api/routers/experiments.router.ts
```diff
@@ -77,6 +79,194 @@ export const experimentsRouter = createTRPCRouter({
...
+    // Anyone can fork an experiment
+    requireNothing(ctx);
```
corbt: We probably should use `requireCanViewExperiment(ctx)` here instead, which is currently the same as `requireNothing(ctx)` but we'll probably let people have totally-private experiments in the future.

src/server/api/routers/experiments.router.ts
```diff
@@ -77,6 +79,194 @@ export const experimentsRouter = createTRPCRouter({
...
+    const maxSortIndex =
+      (
+        await prisma.experiment.aggregate({
+          _max: {
+            sortIndex: true,
+          },
+        })
+      )._max?.sortIndex ?? 0;
+
+    const newExperiment = await prisma.experiment.create({
```
corbt: Let's put this in the transaction as well so they all fail or succeed together. Can pregenerate the experiment ID.

src/server/api/routers/experiments.router.ts
```diff
@@ -77,6 +79,194 @@ export const experimentsRouter = createTRPCRouter({
...
+      }),
+      prisma.templateVariable.createMany({
+        data: templateVariablesToCreate,
+      }),
+    ]);
+
+    return {
+      ...newExperiment,
+      access: {
+        canView: true,
```
corbt: Let's not return these if we aren't using them. In fact I'd probably just return the new experiment id since that's all we need.

src/components/experiments/HeaderButtons/useOnForkButtonPressed.tsx
arcticfly: No, I think you're right, we don't need it

src/components/experiments/HeaderButtons/useOnForkButtonPressed.tsx
arcticfly: Nah, figured I'd let the users test it out for us. Saves engineering time.

src/pages/experiments/[id].tsx
Comment (arcticfly) [reply src/pages/experiments/[id].tsx]: Yes, I'll move it back where you put it to make the diff cleaner

src/server/api/routers/experiments.router.ts
arcticfly: Okay, updated

src/server/api/routers/experiments.router.ts
arcticfly: Okay, it's in there now

src/server/api/routers/experiments.router.ts
arcticfly: Makes sense

src/components/experiments/HeaderButtons/useOnForkButtonPressed.tsx
arcticfly: Lol yes I checked

PR #95, opened 2023-07-26
Store multiple ModelResponses

arcticfly:
This PR renames `ModelOutput` to `ModelResponse` and stores things like status codes and retry times in the `ModelResponse` rows. It also allows one `ScenarioVariantCell` to have multiple `ModelResponse`s, which in turn allows us to log multiple API requests to the user.

## Changes
* Rename `ModelOutput` to `ModelResponse`
* Store status codes and error messages on the responses
* Change logic for getting model output throughout the app
* Show logs while waiting for cell output to load

Logs:
Screenshot 2023-07-25 at 5 49 32 PM


src/server/api/routers/promptVariants.router.ts
```diff
@@ -51,7 +52,9 @@ export const promptVariantsRouter = createTRPCRouter({
           id: true,
         },
         where: {
-          modelOutput: {
+          modelResponse: {
+            outdated: false,
+            output: { not: Prisma.AnyNull },
```
corbt, line 57: I think this is probably right but did you run an eval or two and make sure the results look reasonable?

corbt (approved):
This will make things feel so much better!

src/server/api/routers/promptVariants.router.ts
arcticfly: Evals appear to be broken already, so I'll debug that in another PR.

PR #99, opened 2023-07-26
Requeue rate-limited query model tasks

arcticfly:
Requeueing query tasks (instead of sleeping within the worker and trying later) allows non-rate-limited tasks to supersede their precedence and complete while the rate limit is being recovered from.

## Changes
* Requeue tasks
* Run evals in background task
* Continue polling VariantStats while evals are running

### DIFF
--- a/src/server/tasks/defineTask.ts
+++ b/src/server/tasks/defineTask.ts
@@ -7,9 +7,9 @@ function defineTask<TPayload>(
   taskIdentifier: string,
   taskHandler: (payload: TPayload, helpers: Helpers) => Promise<void>,
 ) {
-  const enqueue = async (payload: TPayload) => {
+  const enqueue = async (payload: TPayload, runAt?: Date) => {
     console.log("Enqueuing task", taskIdentifier, payload);
-    await quickAddJob({ connectionString: env.DATABASE_URL }, taskIdentifier, payload);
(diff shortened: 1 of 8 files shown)


src/server/tasks/defineTask.ts
```diff
@@ -7,9 +7,9 @@ function defineTask<TPayload>(
   taskIdentifier: string,
   taskHandler: (payload: TPayload, helpers: Helpers) => Promise<void>,
 ) {
-  const enqueue = async (payload: TPayload) => {
+  const enqueue = async (payload: TPayload, runAt?: Date) => {
     console.log("Enqueuing task", taskIdentifier, payload);
-    await quickAddJob({ connectionString: env.DATABASE_URL }, taskIdentifier, payload);
+    await quickAddJob({ connectionString: env.DATABASE_URL }, taskIdentifier, payload, { runAt });
```
corbt, line 12: This works if you enqueue without setting `runAt` right?

corbt (approved):
Seems reasonable. Probably want to rearchitect at some point so older tasks get processed first instead of just whichever one happens to be requeued next but w/e

src/server/tasks/defineTask.ts
arcticfly: Yes

PR #100, opened 2023-07-31
Add og image for experiments

arcticfly:
## Changes
* Add stats endpoint for experiments
* Add `pages/api/experiments/og-image.tsx` route
* Save Inconsolata font ttf in public folder
* Add default open graph tags for site

OG image for experiment:
image


src/pages/_app.tsx
```diff
@@ -21,6 +21,11 @@ const MyApp: AppType<{ session: Session | null }> = ({
           name="viewport"
           content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0"
         />
+        <meta name="og:title" content="OpenPipe" />
+        <meta name="og:description" content="Tinker with the latest LLMs." />
```
corbt: Maybe we should match the tagline you put on our landing page?

Title: "OpenPipe: Open-Source Lab for LLMs"
Description: "OpenPipe is a powerful playground for quickly optimizing performance, cost, and speed across models."

I think I like those better.

arcticfly: Updated

src/pages/experiments/[id].tsx
```diff
@@ -22,13 +25,43 @@ import { useExperiment, useHandledAsyncCallback } from "~/utils/hooks";
 import { useAppStore } from "~/state/store";
 import { useSyncVariantEditor } from "~/state/sync";
 import { HeaderButtons } from "~/components/experiments/HeaderButtons/HeaderButtons";
+import Head from "next/head";
+import { appRouter } from "~/server/api/root.router";
+import { createInnerTRPCContext } from "~/server/api/trpc";
+
+export const getServerSideProps = async (context: GetServerSidePropsContext<{ id: string }>) => {
```
Comment (corbt) [inline src/pages/experiments/[id].tsx:32]: Worth keeping in mind that this will slow down our pageload because now we have to pull the stats on the server before sending down the page and pulling everything else on the client, instead of pulling everything in parallel.

src/pages/experiments/[id].tsx
```diff
@@ -62,53 +95,66 @@ export default function Experiment() {
...
-    </AppShell>
+    <>
+      {stats && (
+        <Head>
+          <meta property="og:title" content={stats.experimentLabel} />
+          <meta
+            property="og:image"
+            content={`/api/experiments/og-image?experimentLabel=${stats.experimentLabel}&variantsCount=${stats.promptVariantCount}&scenariosCount=${stats.testScenarioCount}`}
+          />
+          <meta property="og:image:height" content="630" />
```
Comment (corbt) [inline src/pages/experiments/[id].tsx]: You can use the `key` property on all of these to make it so we don't have duplicate tags in the rendered HTML, which might confuse some processors. https://nextjs.org/docs/pages/api-reference/components/head

corbt (approved):
Let's see what happens

src/pages/experiments/[id].tsx
Comment (arcticfly) [reply src/pages/experiments/[id].tsx]: Good idea

PR #112, opened 2023-08-02
Improve posthog

arcticfly:
## Changes
* Identify posthog users
* Rewrite posthog events

We should add `NEXT_PUBLIC_HOST` to our production server as well.


corbt (approved):

src/utils/analytics.ts
```diff
@@ -1,13 +1,21 @@
 // Make sure we're in the browser
```
corbt: Can we just delete this file and move it all to SessionIdentifier?

src/utils/analytics.ts
```diff
@@ -1,13 +1,21 @@
...
     posthog.init(env.NEXT_PUBLIC_POSTHOG_KEY, {
-      api_host: "https://app.posthog.com",
+      api_host: `${env.NEXT_PUBLIC_HOST}/ingest`,
     });
   }
 }
+
+export const identifySession = (session: Session) => {
+  if (!enableAnalytics) return;
+  posthog.identify(session.user?.id, {
```
corbt: What happens if `user` is null? Should we check for that and not call `identify` in that case?

arcticfly: Makes sense

src/utils/analytics.ts
arcticfly: I'll delete SessionIdentifier and move it in here, since we'll likely add more helper functions later.

PR #120, opened 2023-08-05
Rename 'anthropic' to 'anthropic/completion'

corbt:
More consistency in the way we name our model providers.


arcticfly:
Looks good, should work!

PR #123, opened 2023-08-08
Add Logged Calls and projects

corbt:
Adds the concept of a logged call, which will let users import data from their apps directly.

Also adds the idea of "projects" that can be used to organize experiments and logged calls. We'll also allow you to add collaborators to projects eventually.


app/prisma/migrations/20230808034313_make_model_response_optional/migration.sql
corbt, line 1: When would a LoggedCall not have a ModelResponse?

app/prisma/schema.prisma
```diff
@@ -200,8 +200,11 @@ model DatasetEntry {
     updatedAt DateTime @updatedAt
 }
 
+// TODO rename Organization to Project
 model Organization {
     id                String  @id @default(uuid()) @db.Uuid
+    name              String  @default("Project 1")
```
corbt, line 206: Is there a good reason to set a default here? Most of the time "Project 1" won't be the right name, and I think it's better for us to get a typescript error if we create a project and forget to populate this field than have it have a misleading default.

app/src/components/CopiableCode.tsx
```diff
@@ -0,0 +1,43 @@
+import { HStack, Icon, IconButton, Tooltip, Text } from "@chakra-ui/react";
+import { useCallback, useState } from "react";
+import { MdContentCopy } from "react-icons/md";
+
+const CopiableCode = ({ code }: { code: string }) => {
+  const [copied, setCopied] = useState(false);
+
+  const copyToClipboard = useCallback(() => {
+    const onCopy = async () => {
```
corbt: [Is](url) this inner function just to avoid the lint errors if you have an async `useCallback`? If so why not use `useHandledAsyncCallback`?

app/src/components/CopiableCode.tsx
```diff
@@ -0,0 +1,43 @@
...
+      console.log("copied!");
+      await navigator.clipboard.writeText(code);
+      setCopied(true);
+    };
+    void onCopy();
```
corbt: what is the `void` identifier for? I'm not familiar with that pattern.

app/src/utils/hooks.ts
```diff
@@ -132,3 +149,8 @@ export const useScenario = (scenarioId: string) => {
 };
 
 export const useVisibleScenarioIds = () => useScenarios().data?.scenarios.map((s) => s.id) ?? [];
+
+export const useSelectedOrg = () => {
```
corbt, line 153: (it definitely would have been easier to rename org to project before writing all the additional code in this PR 😂)

app/src/components/nav/ProjectBreadcrumbContents.tsx
```diff
@@ -0,0 +1,34 @@
...
+  orgName = orgName || selectedOrg?.name || "";
+
+  return (
+    <Link href="/home">
+      <HStack w="full">
+        <Flex
+          p={1}
+          borderRadius={4}
+          backgroundColor="orange.100"
+          minW={6}
```
corbt: I am confused don't these 6 values together do the same thing as `boxSize={6}`?

app/src/pages/settings/index.tsx
```diff
@@ -0,0 +1,142 @@
+import {
```
corbt: I think these should live at `/projects/[id]/settings`, or maybe `/projects/settings`? We'll have user (and actual org) settings as well in the future.

app/src/server/api/routers/externalApi.router.ts
```diff
@@ -0,0 +1,205 @@
...
+      if (!apiKey) {
+        throw new TRPCError({ code: "UNAUTHORIZED" });
+      }
+      const key = await prisma.apiKey.findUnique({
+        where: { apiKey },
+      });
+      if (!key) {
+        throw new TRPCError({ code: "UNAUTHORIZED" });
+      }
+      const reqPayload = await reqValidator.spa(input.reqPayload);
```
corbt, line 135: I'm still concerned about this. If someone messes up their prompt construction and accidentally calls OpenAI with `messages` set to `null`, my understanding is this will just throw a server error they'll never see and the log will never show up. Wouldn't it be better to log the request anyway, along with the error response they got from OpenAI, so they know what went wrong?

app/src/server/api/routers/organizations.router.ts
```diff
@@ -0,0 +1,146 @@
...
+        },
+      },
+      orderBy: {
+        createdAt: "asc",
+      },
+    });
+
+    if (!organizations.length) {
+      const newOrgId = uuidv4();
+      const [newOrg] = await prisma.$transaction([
```
corbt: Is there a reason this really has to exist in the list call? We're going to end up with race conditions where multiple projects are created bc of server-side rendering immediately followed by client-side rendering. If it absolutely has to happen is there any way to do it with an upsert?

app/src/server/api/routers/organizations.router.ts
```diff
@@ -0,0 +1,146 @@
...
+          },
+        }),
+      ]);
+      organizations.push(newOrg);
+    }
+
+    return organizations;
+  }),
+  get: protectedProcedure.input(z.object({ id: z.string() })).query(async ({ input, ctx }) => {
+    requireNothing(ctx);
```
corbt: Shouldn't this be protected?

app/src/server/utils/generateApiKey.ts
```diff
@@ -0,0 +1,11 @@
+const KEY_LENGTH = 42;
+
+export const generateApiKey = () => {
+  const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  let randomChars = "";
+  for (let i = 0; i < KEY_LENGTH; i++) {
+    randomChars += chars.charAt(Math.floor(Math.random() * chars.length));
```
corbt: Can we use a dedicated lib for this? There have got to be a bunch and `Math.random` isn't actually cryptographically secure.

client-libs/typescript/openai-legacy/index.ts
```diff
@@ -0,0 +1,81 @@
...
+// Anything we don't override we want to pass through to openai directly
+export * as openAILegacy from "openai-legacy";
+
+type OPConfigurationParameters = {
+  apiKey?: string;
+  basePath?: string;
+};
+
+export class Configuration extends openai.Configuration {
+  public qkConfig?: openPipeClient.Configuration;
```
corbt, line 14: lol probably shouldn't be `qkConfig` anymore. 😂

client-libs/typescript/openai-legacy/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  constructor(config: openai.ConfigurationParameters & { opParameters?: OPConfigurationParameters }) {
```
corbt, line 16: I think naming the arg just `openpipe` instead of `opParameters` will be clearer.

client-libs/typescript/openai/index.ts
```diff
@@ -0,0 +1,111 @@
+import * as openai from "openai-beta";
+import { readEnv, type RequestOptions } from "openai-beta/core";
+import { CompletionCreateParams } from "openai-beta/resources/chat/completions";
+import axios from "axios";
+
+export * as openai from "openai-beta";
+import * as openPipeClient from "../codegen";
+
+interface ClientOptions extends openai.ClientOptions {
+  openPipeApiKey?: string;
```
corbt, line 10: I think as a caller I'd rather just have a single `openpipe` top-level arg everywhere that our stuff nests inside of.

corbt (commented):
This is a really impressive amount to get done! Have some suggestions/comments but overall very happy with where you got to!

app/prisma/migrations/20230808034313_make_model_response_optional/migration.sql
arcticfly: Before I this change took effect, each LoggedCall required a LoggedCallModelResponse, and vice versa. Thus no matter in what order I created the two records, one always failed on a foreign key violation, even within the same transaction. The best solution I could find was to make LoggedCall's relationship to a LoggedCallModelResponse optional and put these 3 operations in the same transaction:

1. Create LoggedCall
2. Create LoggedCallModelResponse, related to newly created LoggedCall
3. Update LoggedCall to relate to the newly created LoggedCallModelResponse

app/prisma/schema.prisma
arcticfly: I think you're right, we don't really need a default name here. This was the simplest way I could think of to name all the existing organization records that don't currently have a name. What do you think a better approach is?

app/src/components/CopiableCode.tsx
arcticfly: Good idea! That's much cleaner.

app/src/components/CopiableCode.tsx
arcticfly: It allows us to avoid awaiting an asynchronous function without any eslint errors. I don't think it used to be necessary, so I must have updated my linter recently.

app/src/utils/hooks.ts
arcticfly: Yes, you're right. I've been thinking about orgs and projects for the last few days, and I think we may need both. Billing should probably be associated with an org, along with membership and privileges. However, I don't think we should invest time setting that up right now.

app/src/components/nav/ProjectBreadcrumbContents.tsx
arcticfly: Yes they do! I didn't realize boxSize could be used on anything besides icons.

app/prisma/schema.prisma
corbt: I usually do this in 3 steps -- create the field as nullable, set the value for all existing entries, then add the non-null constraint. Eg. https://github.com/OpenPipe/OpenPipe/blob/main/app/prisma/migrations/20230718201303_add_users_and_orgs/migration.sql#L79-L86

You could also do it in two steps I suppose -- create the field with a default value, then remove the default value. Seems slightly less clear to me but either way should work.

app/src/pages/settings/index.tsx
arcticfly: Makes sense!

app/src/components/CopiableCode.tsx
corbt: I actually like the fact that eslint tells us if we forget to report an unhandled promise exception -- keeps us from silently swallowing errors accidentally.

app/src/server/api/routers/externalApi.router.ts
arcticfly: Oh I see what you mean. Yes, let's change this to no longer require proper validation.

app/src/server/api/routers/organizations.router.ts
arcticfly: No it doesn't need to exist in the list call. This is merely a hack to ensure the user automatically gets a project after signing up. A better solution would be to add a new route which specifically checks if the user doesn't already have a project and upserts one in that event. We should then call that route every time a user signs up. I'm not sure how to do that with github auth, but it's probably possible.

app/src/server/api/routers/organizations.router.ts
arcticfly: Yes it should (and now is)

app/src/server/utils/generateApiKey.ts
arcticfly: We are now using `crypto-random-string`

client-libs/typescript/openai-legacy/index.ts
arcticfly: Yeah, we should change that

client-libs/typescript/openai-legacy/index.ts
arcticfly: Okay, makes sense to me!

client-libs/typescript/openai/index.ts
arcticfly: Yes, I agree

app/src/pages/settings/index.tsx
arcticfly: It's at `/project/settings` now

PR #148, opened 2023-08-12
Fix client bugs

corbt:
1. PostHog can only be used client-side
2. Can't nest <a> tags in the ProjectMenu
3. Don't re-load Monaco on every render cycle (oops 😂)


arcticfly:
> 3\. Don't re-load Monaco on every render cycle (oops 😂)

😆

arcticfly:
> 1. PostHog can only be used client-side

That only refers to the posthog client package, right? I think we're still using `capturePath` in trpc.ts, which uses `posthog-node`

corbt:
> That only refers to the posthog client package, right? I think we're still using capturePath in trpc.ts, which uses posthog-node


Yep that's right, good clarification.

PR #149, opened 2023-08-12
Load the JS client using pnpm workspaces

corbt:
This makes it so we're using our own openpipe client for all OpenAI calls from the OpenPipe app.

The client doesn't do anything at the moment beyond proxying to the OpenAI lib. But this infra work should make it easier to quickly iterate on the client and test the changes in our own app.


client-libs/typescript/src/openai-legacy/index.ts
```diff
@@ -0,0 +1,90 @@
+// import * as openPipeClient from "../codegen";
```
corbt, line 1: Commented this out because I was getting typescript errors, will need to re-enable it once we get to trying to implement.

arcticfly:
> This makes it so we're using our own openpipe client for all OpenAI calls from the OpenPipe app.
> 
> The client doesn't do anything at the moment beyond proxying to the OpenAI lib. But this infra work should make it easier to quickly iterate on the client and test the changes in our own app.

Nice work! What's preventing us from capturing metrics through our typescript library in production? It's already working locally.

corbt:
> Nice work! What's preventing us from capturing metrics through our typescript library in production? It's already working locally.

Not sure; it wasn't working immediately and I didn't want to block the infra work on getting the library actually working.

PR #152, opened 2023-08-13
Create non-functional experiment from request logs

arcticfly:
The user can now select some logs and create an experiment which assumes that all original logs were produced by GPT 3.5 and that the user wants to compare the original responses to those that GPT 4 produces. We currently have no safe-guards to ensure that all original requests were made to the same model, and no way to allow the user to select the model they wish to try out. Assuming that we get the original and new variant models correct, there is still no way to produce new output because our current architecture requires us to run the experiment once with empty scenario variable values, and that causes our type checking to fail. Still, this PR lays the groundwork for creating experiments from logged calls.

Screenshot 2023-08-13 at 3 25 59 AM


arcticfly:
We'll actually do this through a dataset

PR #158, opened 2023-08-15
Filter logged calls

arcticfly:
Users can now filter their logs by `Request`, `Response`, `Model`, and `Status Code`, along with any tag they've added to it. Comparators are currently limited to `=`, `!=`, and `CONTAINS`. Tagged values now appear in the logged calls table.

Screenshot 2023-08-15 at 2 57 02 AM


app/package.json
```diff
@@ -10,13 +10,13 @@
   },
   "scripts": {
     "build": "next build",
-    "dev:next": "next dev",
+    "dev:next": "TZ=UTC next dev",
```
arcticfly, line 13: There may be a more elegant way to do this. For some reason Kysely gets all confused about timezones if we don't inform it that our dates are stored in UTC time.

app/src/components/requestLogs/LogFilters/LogFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+import SelectComparatorDropdown from "./SelectComparatorDropdown";
+
+const LogFilter = ({ filter, index }: { filter: LogFilter; index: number }) => {
+  const updateFilter = useAppStore((s) => s.logFilters.updateFilter);
+  const deleteFilter = useAppStore((s) => s.logFilters.deleteFilter);
+
+  const [editedValue, setEditedValue] = useState("");
+
+  const debouncedUpdateFilter = useCallback(
+    debounce(
```
arcticfly, line 18: Debouncing to save our poor database

corbt: lol debouncing sounds good, but maybe we want >200ms? We might actually need to wait until someone clicks "save filters" or something. This is likely going to slam us once a user has have a non-trivial number of logs (more than a few hundred, even).

client-libs/typescript/src/openai/mergeChunks.ts
```diff
@@ -1,8 +1,9 @@
 import { ChatCompletion, ChatCompletionChunk } from "openai-beta/resources/chat";
+import { omit } from "lodash-es";
```
corbt, line 2: Ouch, sorry about that. I had a @ts-expect-error for another problem on the relevant line that was masking the missing import!

corbt (approved):
This is really good. The filters logic is kinda hairy and there might be a clever way to simplify it some, but what you've got right now is easy enough to read and it's not immediately obvious to me how to do it better anyway.

app/package.json
corbt: Huh. What are the symptoms of it getting confused about timezones? Does it return timestamps in the wrong timezone so they display wrong to the user?

corbt:
Going to go ahead and merge this since the missing "lodash-es" dep is breaking openai completions in prod.

app/package.json
arcticfly: Kysely very helpfully converts the times from PT to UTC for us (that is, it adds 7 hours to whatever time is stored in the database)

PR #161, opened 2023-08-16
Invite members

arcticfly:
Project admins can now invite other members to join their projects. We send an invitation email through Brevo (previously Send In Blue) that contains an invitation token, with which any authenticated user can join the associated project.

Screenshots:
Screenshot 2023-08-16 at 2 37 15 AM
Screenshot 2023-08-16 at 2 36 20 AM
Screenshot 2023-08-16 at 2 36 39 AM
Screenshot 2023-08-16 at 2 38 09 AM
Screenshot 2023-08-16 at 2 38 24 AM


app/src/components/projectSettings/InviteMemberModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  RadioGroup,
+  Radio,
+} from "@chakra-ui/react";
+import { useState, useEffect } from "react";
+
+import { api } from "~/utils/api";
+import { useHandledAsyncCallback, useSelectedProject } from "~/utils/hooks";
+import { maybeReportError } from "~/utils/errorHandling/maybeReportError";
+
+type ProjectRole = "MEMBER" | "ADMIN";
```
corbt: FYI I think we can import this type from Prisma

app/src/server/emails/sendEmail.ts
```diff
@@ -0,0 +1,34 @@
+import * as SibApiV3Sdk from "@sendinblue/client";
```
corbt: I probably would have used a generic SMTP email sender for this instead of using Brevo's API directly. That way it's easy for us to swap out for another provider in local dev for testing.

Eg. this is how I set up emails with Brevo in another project:

```typescript
import { marked } from "marked";
import nodemailer from "nodemailer";
import { env } from "~/env.mjs";

const transporter = nodemailer.createTransport({
  // All the SMTP_ env vars come from https://app.brevo.com/settings/keys/smtp
  // @ts-expect-error nodemailer types are wrong
  host: env.SMTP_HOST,
  port: env.SMTP_PORT,
  auth: {
    user: env.SMTP_LOGIN,
    pass: env.SMTP_PASSWORD,
  },
});

export const sendEmail = async (options: { to: string; subject: string; body: string }) => {
  const bodyHtml = marked.parseInline(options.body, { mangle: false });

  return await transporter.sendMail({
    from: "noreply@translato.ai",
    to: options.to,
    subject: options.subject,
    html: bodyHtml,
    text: options.body,
  });
};
```

corbt (approved):
Looks good!

app/src/server/emails/sendEmail.ts
arcticfly: Good idea

PR #163, opened 2023-08-17
Add openpipe/Chat provider with Open-Orca/OpenOrcaxOpenChat-Preview2-13B model

arcticfly:
## Changes
* Add `openpipe/Chat` provider
* Add `Open-Orca/OpenOrcaxOpenChat-Preview2-13B` as its first model
* Add `templatePrompt` as an optional field on frontend models


corbt (approved):
Seems reasonable enough!

PR #207, opened 2023-08-31
Add docs

arcticfly:
Add the necessary docs for our waitlisters to get up and running with the SDK.

Screenshot 2023-08-31 at 3 22 17 AM


corbt:
Just looking at the screenshot, I'd make the intro blurb much less salesy. It should be a straightforward description of what developers can use OpenPipe to do and maybe give a very high-level sense of how the system works from a user's perspective. Devs really don't like sales language ime.

docs/overview.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: "Overview"
+description: "OpenPipe is a streamlined platform designed to help product-focused teams train specialized LLM models as replacements for slow and expensive prompts."
+---
+
+## Who We Are
+
+We're a team of full-stack engineers and machine learning researchers who came together to streamline the process of integrating fine-tuned models into any application. We believe that specialized models will continue to outperform enormous multipurpose LLMs for decades to come, and our goal is to make the fine-tuning process accessible to everyone.
```
corbt: Don't think docs are the right place to put this kind of stuff. Devs are turned off by stuff that sounds like marketing. This should focus more on what the experience of using openpipe is like from a dev's point of view, not our Big Vision

PR #213, opened 2023-09-04
Add useful datasets

arcticfly:
## Changes
* Save input/output on datasets
* Add ability to edit input/output
* Add datasets tab

Screenshot 2023-09-05 at 11 23 23 AM


corbt:
A few initial questions here:

1. Should we assign a train/test split to dataset entries at the time they're added to the dataset? I think probably yes.
2. It looks like fine-tuning is still kicked off from the logged calls page instead of from the datasets page. That should probably be moved to the dataset page? If we do want to keep a "fine-tune" button on the logged call page (pros/cons there) it should probably *first* generate a dataset in the background and then use that dataset for training.
3. Need to have some way to track what exact data went into a model. I'm not sure how to do this and I guess we can punt for now, but seems pretty important long-term for debugging, data provenance and measuring different training runs.

app/src/components/OutputsTable/VariantEditor.tsx
```diff
@@ -132,6 +132,7 @@ export default function VariantEditor(props: { variant: PromptVariant }) {
         wordWrapBreakBeforeCharacters: "",
         quickSuggestions: true,
         readOnly: !canModify,
+        scrollBeyondLastLine: false,
```
corbt: I'm worried that if there's a long last line it could get hidden behind the save button with no way to see it? Also I like the default behavior that lets you scroll past the code, is there a strong reason to remove it?

arcticfly: I keep accidentally scrolling the code editor when I intend to scroll the screen, which ends with me staring at a few syntax characters, like this:

Screenshot 2023-09-04 at 5 06 47 PM

However, I don't feel that strongly, so I'll revert this change in the VariantEditor. I do want to disable scrolling beyond the last line for the JSON editor in the dataset entry drawer.

arcticfly:
> A few initial questions here:
> 
> 1. Should we assign a train/test split to dataset entries at the time they're added to the dataset? I think probably yes.
> 2. It looks like fine-tuning is still kicked off from the logged calls page instead of from the datasets page. That should probably be moved to the dataset page? If we do want to keep a "fine-tune" button on the logged call page (pros/cons there) it should probably _first_ generate a dataset in the background and then use that dataset for training.
> 3. Need to have some way to track what exact data went into a model. I'm not sure how to do this and I guess we can punt for now, but seems pretty important long-term for debugging, data provenance and measuring different training runs.

1. We ended up talking about this and deciding that we would statistically assign a type to an entry as it comes in.
2. I moved fine-tuning to only be kick-offable from the dataset page.
3. What we're doing right now is associating a dataset with a model. That isn't a perfect solution, since the dataset (and especially what part of it is training/testing) can change, but it will allow us a good idea of the type of data used for a model. One potential solution here would be to create a new dataset for each fine-tune, which would bloat our database but provide a nice history.

PR #222, opened 2023-09-12
Add support for CommonJS in sdk

arcticfly:
Currently, our https://npmjs.org/openpipe typescript sdk is only compatible with ECMAScript projects. Since one of our customers uses CommonJS, this PR introduces a tsup build step that bundles all of our typescript code to be compatible with both ECMAScript and CommonJS imports.

## Changes
* Use tsup to generate dist folder
* Move sdk source code into /src folder
* Generate cjs files for all entry points


corbt:
Seems plausible. How are you checking that this actually works in different consumer app setups?

arcticfly:
> Seems plausible. How are you checking that this actually works in different consumer app setups?

I've set up a test project that uses our repo as a yalc dependency in https://github.com/OpenPipe/test-projects. We should add tests for python as well.

client-libs/typescript/tsup.config.ts
```diff
@@ -0,0 +1,25 @@
...
+  splitting: false, // Disable code splitting
+  sourcemap: true, // Include sourcemaps
+  target: "es2020", // Target ES2020 syntax for modern environments
+  dts: true, // Generate declaration files
+
+  // Define entry points
+  entry: [
+    "src/index.ts", // Main entry
+    "src/openai.ts", // 'openai' sub-module
+    "src/openai/index.test.ts", // 'openai' sub-module
```
corbt: Why does this need to be an entrypoint? Don't we just need src/index and src/openai?

arcticfly: Yes, you are correct, I thought I had removed that.

PR #228, opened 2023-09-13
Update readme with Github Callback URL example

### LINKED ISSUE #225: Sweep: Github oauth app - callback url? / constantly redirecting
### Details

I just tried setting up OpenPipe as per the documentation, I created the Github oauth app but wasn't sure what to put in the callback URL - should this just be your local install base URL?

I suspect this is what is causing my instance to load -> redirect to github (and auth fine) -> redirect back to my app -> redirect back to github etc... over and over forever.

sammcj:
Fixes #225


arcticfly:
Thanks @sammcj

PR #252, opened 2023-09-23
Add internal router and trainer

arcticfly:
This PR allows our training docker containers to get a shared access signature url from the main server to download a fine-tuned model's training data from Azure.

### Changes
* Add `BaseModel` enum, `~utils/baseModel.ts`
* Add `codegen:internalClient` pnpm command
* Add `api/internal/v1` router with `getTrainingInfo` route
* Add docker client with basic test
* Add `generateBlobDownloadUrl` function


app/.env.example
```diff
@@ -13,6 +13,9 @@
 # https://www.prisma.io/docs/reference/database-reference/connection-urls#env
 DATABASE_URL="postgresql://postgres:postgres@localhost:5432/openpipe?schema=public"
 
+# Used by docker containers to request training data from the API
+DOCKER_SECRET="placeholder"
```
corbt: This isn't a great name. For one thing, all our software runs in Docker. Maybe AUTHENTICATED_SYSTEM_KEY?

app/src/pages/api/internal/v1/openapi.json.ts
```diff
@@ -0,0 +1,16 @@
+import { type NextApiRequest, type NextApiResponse } from "next";
+import { generateOpenApiDocument } from "trpc-openapi";
+import { v1ApiRouter } from "~/server/api/internal/v1Api.router";
+
+export const openApiDocument = generateOpenApiDocument(v1ApiRouter, {
+  title: "OpenPipe Docker API",
```
corbt: "OpenPipe Internal API". Docker isn't a useful description. We may or may not use Docker as the runtime, and anyway, our main app runs in Docker too, so it's a weird thing to use as a distinguisher.

app/src/server/api/internal/openApiTrpc.ts
```diff
@@ -0,0 +1,45 @@
...
+          ...shape.data,
+          zodError: error.cause instanceof ZodError ? error.cause.flatten() : null,
+        },
+      };
+    },
+  });
+
+export const createOpenApiRouter = t.router;
+
+export const openApiPublicProc = t.procedure;
```
corbt: Shouldn't this be `openApiProtectedProc`?

docker-client/openapi.json
```diff
@@ -0,0 +1,118 @@
+{
```
corbt: I don't see any code to generate this file, did you just manually copy/paste it? If we aren't autogenerating this file probably better not to have it at all and run `poetry run openapi-python-client generate --url http://localhost:3000/api/internal/openapi.json` or whatever.

corbt (changes requested):

docker-client/openapi.json
corbt: Nvm I see where you generate this file now.

arcticfly:
Updated the PR to use the word "Docker" in fewer places. Not sure why, but I was really focused on running this job in a docker container when I devised the folder structure and variable names!

corbt (approved):

PR #255, opened 2023-09-26
Make dataset entries immutable, snapshot fine-tune training entries and pruning rules

arcticfly:
This PR allows us to record what data a fine-tuned model was trained on, and to ensure its pruning rules aren't altered after it is created. We should make sure to run `npx tsx src/server/scripts/backfillTrainingDatasets.ts` once this is merged to backfill the training entry information for existing fine-tuned models.

### Changes
* Add `FineTuneTrainingEntry` to record what entries were used to train a model
* Add `outdated`, `persistentId` and `sortKey` to dataset entries
* Hide `outdated` entries in the default dataset view
* Record the user who altered a given version of a dataset entry
* Allow pruning rules to be associated with a fine-tuned model instead of a dataset
* Add `backfillTrainingDatasets.ts` script


app/src/server/scripts/backfillTrainingDatasets.ts
```diff
@@ -0,0 +1,50 @@
+import { type Prisma } from "@prisma/client";
```
corbt, line 1: I probably would have explained what we want to do here and had GPT-4 do this using SQL in the migration file so we don't have to run another script. But not worth redoing for sure. 🙂

app/src/server/api/routers/fineTunes.router.ts
```diff
@@ -87,8 +84,58 @@ export const fineTunesRouter = createTRPCRouter({
...
+                  matches: true,
+                },
+              },
+            },
+          },
+        },
       });
       if (!fineTune) return error("Error creating fine tune");
+
+      await prisma.fineTuneTrainingEntry.createMany({
```
corbt, line 114: This could be a little bit expensive for larger datasets bc the matches arrays could get really big. I'd probably have leaned towards doing this in the background job instead, but we can leave it here and see whether it causes problems.

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -260,20 +264,42 @@ export const datasetEntriesRouter = createTRPCRouter({
...
+          input: parsedInput ?? prevEntry.input,
+          output: parsedOutput ?? prevEntry.output,
+          inputTokens: inputTokens ?? prevEntry.inputTokens,
+          outputTokens: outputTokens ?? prevEntry.outputTokens,
+          type: input.updates.type ?? prevEntry.type,
+          datasetId: prevEntry.datasetId,
+          sortKey: prevEntry.sortKey,
+          authoringUserId: ctx.session?.user.id,
+          persistentId: prevEntry.persistentId,
+          matchedRules: {
```
corbt, line 292: I think we need to recalculate matched rules on every update?

corbt (approved):
Great work. Only open question for me is what happens if an update causes the matched rules to change? Seems like we don't account for that here.

app/src/server/api/routers/fineTunes.router.ts
arcticfly: We don’t create fine-tunes very often, so I’m okay with it being slow.

app/src/server/api/routers/datasetEntries.router.ts
arcticfly: That's what we're doing on line 300 in `updatePruningRuleMatches`.

PR #266, opened 2023-10-03
Show fine-tuned model output on test cases

arcticfly:
Users will now be able to see the output of their fine-tuned model when fed test cases as inputs.

### Changes
* Add `FineTuneTestingEntry` to schema
* Add `getTestResult` task, queue it from the `checkFineTuneStatus` task
* Add view for test case output to fine-tuned model page
* Use NextJS url library to save query parameters, save selected model tab in url
* Fix client code generation
* Prevent user from saving dataset entry without changes


app/src/server/tasks/fineTuning/checkFineTuneStatus.task.ts
```diff
@@ -22,7 +27,40 @@ const runOnce = async () => {
       try {
         const resp = await trainingStatus({ callId: job.modalTrainingJobId });
         if (resp.status === "done") {
-          // TODO: kick off job to run test set evals here
+          const fineTune = await prisma.fineTune.findUnique({
+            where: { id: job.id },
+          });
+          if (!fineTune) return;
+          const pruningRules = await prisma.pruningRule.findMany({
```
corbt: EDIT: looks like the `pruneInputMessages` function does some of the abstraction I was referring to here. But I still think it needs to order the messages so they're applied in a consistent way?

This needs to be extracted and shared between `getCompletion2` and here at least. Also probably with the script that uploads training data. It's way too easy to update the way pruning rules are applied in one place and not another and break things.

Separately, I think there's a potential bug here? Don't pruning rules need to be applied in the same order they're defined or else the results won't necessarily be stable? It doesn't look like this code sorts them before applying them. Haven't checked if we do so in other places but if not I think we need to.

app/src/components/datasets/DatasetEntriesTable/DatasetEntryEditorDrawer.tsx
```diff
@@ -57,6 +58,15 @@ export default function DatasetEntryEditorDrawer({
     }
   }, [savedInputMessages, savedOutputMessage]);
 
+  const initialHash = useMemo(
```
corbt: I might be missing something but this feels unnecessarily convoluted/difficult to follow? Seems like we could get the same effect in a more straightforward way with something like:

```ts
import { isEqual } from 'lodash-es';
// ...

const hasUpdates = useMemo(
  () =>
    !isEqual(datasetEntry?.input, inputMessagesToSave) ||
    !isEqual(datasetEntry?.output, outputMessageToSave),
  [datasetEntry, inputMessagesToSave, outputMessageToSave],
);
```

app/src/components/fineTunes/ContentTabs/ContentTabs.tsx
```diff
@@ -1,8 +1,10 @@
 import { useState, useRef, useEffect, forwardRef } from "react";
-
 import { Button, VStack, HStack, Text, Divider, Box } from "@chakra-ui/react";
+
 import General from "./General/General";
 import TrainingData from "./TrainingData/TrainingData";
+import TestingData from "./TestingData/TestingData";
```
corbt: Should we rename everything from `TestingData` to `TestSet` to be consistent with what other people call it?

app/src/components/fineTunes/ContentTabs/TestingData/TestingDataRow.tsx
```diff
@@ -0,0 +1,140 @@
...
+      alignItems="flex-start"
+      sx={{
+        td: { verticalAlign: "top" },
+      }}
+    >
+      <Td>
+        <FormattedInput prunedInput={prunedInput} />
+      </Td>
+      <Td
+        sx={{
```
corbt: Just curious, does this not work with inline Chakra properties?

app/src/server/tasks/fineTuning/checkFineTuneStatus.task.ts
arcticfly: > This needs to be extracted and shared between getCompletion2 and here at least. Also probably with the script that uploads training data. It's way too easy to update the way pruning rules are applied in one place and not another and break things.

I'm fine with making the change to load the strings to prune in a standardized function, but that does mean an extra database call in some contexts, for example `getTestResult.task`, since I've been tacking on the pruning rules to existing queries, which is why it isn't currently standardized.

> Separately, I think there's a potential bug here?

You're right, we weren't ensuring correct order here. I'll make that change.

app/src/components/datasets/DatasetEntriesTable/DatasetEntryEditorDrawer.tsx
arcticfly: Yep, forgot about isEqual

app/src/server/tasks/getTestResult.task.ts
```diff
@@ -0,0 +1,174 @@
...
+    return;
+  }
+
+  let shouldRetry = false;
+  let completion;
+  const input = {
+    model: `openpipe:${fineTune.slug}`,
+    messages: datasetEntry.input as unknown as ChatCompletionMessage[],
+  };
+  try {
```
corbt, line 100: Can we extract out the API code and reuse it in both places? Basically I think we should have a function that looks something like `getCompletion(fineTune: FineTune, input: { prompt: string, max_tokens: int, ... }`. That function can figure out the pipeline version, apply the pruning rules in the way the model expects, etc.

Just think it's quite likely we'll forget to this code in both places as we update the pipeline.

app/src/components/fineTunes/ContentTabs/ContentTabs.tsx
arcticfly: I don't feel strongly. I'll rename it.

app/src/utils/hooks.ts
```diff
@@ -105,14 +106,36 @@ export const useElementDimensions = (): [RefObject<HTMLElement>, Dimensions | un
 };
 
 export const usePageParams = () => {
-  const [pageParams, setPageParams] = useQueryParams({
-    page: NumberParam,
-    pageSize: NumberParam,
-  });
+  const router = useRouter();
```
corbt, line 109: Just out of curiousity, what is this fixing?

trainer/openapi.json
```diff
@@ -38,7 +38,10 @@
                 "schema": {
                   "type": "object",
                   "properties": {
-                    "trainingBlobDownloadUrl": {
+                    "trainingDataUrl": {
```
corbt, line 41: I don't think we need this file anymore, the client codegen reads the openapi spec from the dev server on localhost directly.

app/src/components/fineTunes/ContentTabs/TestingData/TestingDataRow.tsx
arcticfly: Inline Chakra properties work fine here. This syntax is an artifact of the process I went through during initial styling, in which I was adding properties to the Tr to be shared across at Td elements. I'll put the styling back in inline Chakra properties for simplicity now.

app/src/server/tasks/getTestResult.task.ts
arcticfly: I don't see a good way to do that because we need to handle models not being set up for inference differently in `getTestResult` and the `createChatCompletion` route.

In the case of `getTestResult`, if the model is not set up correctly, we need to update the testing entry's error message and return. In the case of `createChatCompletion`, we need to throw a TRPCError. In both cases, if an error is caused by anything other than the model not being set up for inference, we need to handle that error differently, meaning that we can't simply create a new `getWrappedCompletion` function that handles pipeline routing and throws an error if the model is not set up for inference, then handle that error the same way we're now handling these model inference set up errors - because if the error that `getWrappedCompletion` throws is not a model setup error, we need to handle that differently.

One option is to create a new `getWrappedCompletion` function that throws a error with the message "The model is not set up for inference" and then specifically check for that error message in whatever error is thrown by `getWrappedCompletion`, and handle it the way we currently do in these separate calling contexts.

app/src/utils/hooks.ts
arcticfly: Previously we were overwriting all query params (meaning setting page number would remove the active tab param). When I stopped overwriting all params, our code then added `id` (for fine tune id) to the query params, which was long and annoying. NextJS's built-in router handles both these problems well.

trainer/openapi.json
arcticfly: Which file don't we need?

app/src/server/tasks/getTestResult.task.ts
arcticfly: Aside from all that, I don't think we should invest too much time in pipeline-version specific code, because unless I'm mistaken we'll be replacing all version 0 models with version 1 models soon anyway.

app/src/server/tasks/fineTuning/checkFineTuneStatus.task.ts
corbt: If we ensure the `createdAt` and `id` fields are always queried on the pruning rules we could do the sorting within the `pruneInputMessages` function itself. But the way you have it now is fine.

app/src/server/tasks/getTestResult.task.ts
corbt: We'll delete all v0 code, but we'll always need pipeline management code because at some point (likely soon, there are things I think we can improve on the current design) we'll release a v2, v3, etc. and still need to support the old models.

I'd probably have the extracted function return either the completion or an error, and then let the caller decide how to expose the error to the user. But doesn't have to block this PR I guess.

trainer/openapi.json
corbt: I don't think we need `trainer/openapi.json`.

corbt (approved):
Still think we should abstract away the `getCompletion` stuff better but also think getting this out ASAP so Chris can use it is important, so don't have to block on that if it isn't fast to get in.

PR #269, opened 2023-10-04
Calculate score for function outputs

arcticfly:
For models that produce structured data, we now compare the function names and argument values to calculate a per-row score between 0 and 1. We also display the total score at the top of the test set.

### Changes
* Add `caculateEntryScore`, save score on testing entry in database
* Display per-row score and total score in test set tab
* Skip cache when re-running tests for a fine-tuned model

Screenshot 2023-10-03 at 7 22 16 PM

### DIFF
--- a/app/src/server/utils/calculateEntryScore.ts
+++ b/app/src/server/utils/calculateEntryScore.ts
@@ -2,7 +2,7 @@ import { type ChatCompletionMessage } from "openai/resources/chat";
 
 // If function names don't match, return 0
 // If function names match and there are no args, return 1
-// If function names match and there are args, return .5 + (num matching args / num args) * .5
+// If function names match and there are args, return (num matching args / num args)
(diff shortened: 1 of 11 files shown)


app/src/server/utils/calculateEntryScore.ts
```diff
@@ -0,0 +1,41 @@
+import { type ChatCompletionMessage } from "openai/resources/chat";
+
+// If function names don't match, return 0
+// If function names match and there are no args, return 1
+// If function names match and there are args, return .5 + (num matching args / num args) * .5
```
corbt: If the fn names match then we should just return the fraction of args in the output that exactly match the fraction of args in the input. Most of the time there's only one possible fn people are calling so you shouldn't get 50% credit for calling it.

arcticfly: Works for me

PR #271, opened 2023-10-05
Add healthcheck

arcticfly:
This healthcheck will enable Render to restart our service if it becomes unhealthy.


corbt:
You'll also want to update our render.yaml file with the `healthCheckPath`. See https://render.com/docs/blueprint-spec

arcticfly:
> You'll also want to update our render.yaml file with the `healthCheckPath`. See https://render.com/docs/blueprint-spec

Sounds good. I was going to add it in the UI, but render.yaml is better for devs who fork.

Screenshot 2023-10-04 at 6 38 08 PM

PR #288, opened 2023-10-13
Sample selected logs, avoid passing matching log ids in json

arcticfly:
This PR allows the user to sample a selection of selected logs instead of requiring them to create a dataset out of all of them. It also increases performance by avoiding passing matching log ids when querying logs and when creating dataset entries from them. Instead, we calculate which entries the user wants to modify based on `selectedLogIds`, `deselectedLogIds`, and `defaultToSelected` in the selected logs slice.

### Changes
* Add sampling, with 20000 row limit
* Move logic for querying selected logs to utils file, use for filtering logs when creating dataset entries
* Revamp logic in selected logs slice

### DIFF
--- a/app/scripts/run-dev.ts
+++ b/app/scripts/run-dev.ts
@@ -1,6 +1,10 @@
-import { $, ExecaChildProcess } from "execa";
+import { $, type ExecaChildProcess } from "execa";
 import "dotenv/config";
-import humanId from "human-id";
+import HumanIdModule from "human-id";
+
+import { ensureDefaultExport } from "~/utils/utils";
+
+const humanId = ensureDefaultExport(HumanIdModule);
 
 const tunnelSubdomain =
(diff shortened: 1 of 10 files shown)


corbt (approved):
I haven't reviewed the code yet but conceptually the way you described it seems fine.

I probably would have simplified the process (at the cost of flexibility) by copying how Gmail does it: if someone clicks the "select all" checkbox it selects all the visible rows, and pops up a contextual button that says "Select all X,XXX matching entries". If they hit that button then you just move them to the "everything matching the current filters is selected" mode. If they then change the filters, or try to deselect something, then they're snapped out of that mode.

I'm not sure it's worth the code and UX complexity to allow for "select everything matching the filters, BUT also allow manual unselection" -- use cases for that seem pretty narrow, and could be fulfilled by exporting everything to a dataset and then deleting the rows you don't like on that side. But since you have it anyway seems fine to keep.

PR #292, opened 2023-10-16
Add evaluations and model tabs to dataset

arcticfly:
The evaluations tab allows us to compare fine tune outputs against one another.

### Changes
* Add `Model`, `Evaluate`, and `Settings` tabs to dataset page
* Move tab info to url path instead of query parameter
* Remove `Test Set` tab from fine-tune page
* Change default page to be `/request-logs`


corbt:
UX review:
 - there isn't bottom padding on "models" tab
 - there isn't right padding on "evaluate" tab
 - each tab should be its own url eg /[dataset-id]/evaluate not /[dataset-id]?tab=evaluate. If we can keep the animated transition between tabs that's nice but if not getting the URLs right is more important.
 - the fine tune "general" tab needs to have a prominent link to the eval results for just that model.
 - "Settings" should be its own tab, not a drawer that's always present. Right now if I'm on the "Evaluate" tab and click "Settings" I assume I'm looking at eval settings and get confused.
 - On "Evaluate" horizontal scrolling should just scroll the table, not the tabs or header
 - Rename "Output Columns" to "Show Models". Also, this button should be right-justified for consistency with our other pages.
 - There's a weird border radius in the top-left corner of each cell in the header. Should just be straight to the top.

app/src/components/datasets/DatasetContentTabs/Evaluation/ColumnVisibilityDropdown.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const isClientRehydrated = useIsClientRehydrated();
+  if (!isClientRehydrated) return null;
+
+  // If visibleColumns is empty, all columns are visible
+  const numVisibleColumns = visibleColumns.includes(EMPTY_OUTPUT_COLUMNS_KEY)
+    ? 0
+    : visibleColumns.length || columnVisibilityOptions.length;
+
+  console.log("visibleColumns", visibleColumns);
```
corbt: nit rm

app/src/components/datasets/DatasetContentTabs/Evaluation/EvaluationTable/EvaluationTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }, [entries?.deployedFineTunes, visibleColumns]);
+  const numOutputColumns = visibleFineTuneIds.length + (showOriginalOutput ? 1 : 0);
+
+  if (!entries) return null;
+
+  return (
+    <VStack w="full" h="full" justifyContent="space-between" px={8}>
+      <HStack w="full">
+        <Card flex={1} minW="fit-content" variant="outline">
+          <Grid
```
corbt, line 43: lol grids

app/src/utils/hooks.ts
```diff
@@ -138,6 +138,40 @@ export const usePageParams = () => {
   return { page, pageSize, setPageParams };
 };
 
+export const useVisibleEvaluationColumns = () => {
```
corbt: This should definitely live in the `/Evaluation/EvaluationTable` directory. There's no reason to throw everything in one file just because it happens to be a hook.

app/src/utils/hooks.ts
```diff
@@ -138,6 +138,40 @@ export const usePageParams = () => {
...
+  const router = useRouter();
+
+  // Split the "compare" query by commas to get the array of strings.
+  const visibleColumns =
+    typeof router.query.compare === "string"
+      ? router.query.compare.split(",").filter((col) => !!col)
```
corbt: I suspect at this level of complexity we'd be better served by adopting something like https://www.npmjs.com/package/use-query-params but haven't actually used it so I guess I'm not sure it'll make the code simpler/do exactly what we want.

app/src/components/datasets/DatasetContentTabs/Evaluation/EvaluationTable/FineTuneHeading.tsx
```diff
@@ -0,0 +1,69 @@
+import { useState, useEffect } from "react";
+import { Text, VStack, HStack, Tooltip, Box, Icon, GridItem } from "@chakra-ui/react";
+import Link from "next/link";
+import { BsQuestionCircle } from "react-icons/bs";
+
+import ColoredPercent from "~/components/ColoredPercent";
+import { useFineTuneTestingStats, useTestingEntries } from "~/utils/hooks";
+
+const FineTuneHeading = ({ fineTuneId }: { fineTuneId: string }) => {
```
corbt: Nit `FineTuneHeader`

app/src/components/datasets/DatasetContentTabs/Models/Models.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useDatasetFineTunes } from "~/utils/hooks";
+
+const Models = () => {
+  const fineTunes = useDatasetFineTunes().data;
+
+  if (!fineTunes?.length) {
+    return (
+      <Text px={8}>
+        No models have been trained on this dataset. You can start training a model in the{" "}
+        <b>General</b> tab.
```
corbt: Nit: make this a link

corbt (changes requested):

app/src/utils/hooks.ts
arcticfly: We were actually using that package a few weeks ago, but I found that it had an odd interaction with NextJS that caused the [id] portion of a route to be shown as a query param in addition to being shown in the path when we used non-replacing query param updates.

app/src/components/datasets/DatasetContentTabs/Evaluation/EvaluationTable/FineTuneHeading.tsx
arcticfly: Sure I'll make that change, but what's the difference?

arcticfly:
> UX review:
> 
> * there isn't bottom padding on "models" tab

Good catch, added some.

> * there isn't right padding on "evaluate" tab

As we discussed, I ended up adopting the hacky solution on enforcing padding through a div.

> * each tab should be its own url eg /[dataset-id]/evaluate not /[dataset-id]?tab=evaluate. If we can keep the animated transition between tabs that's nice but if not getting the URLs right is more important.

Made that change.

> * the fine tune "general" tab needs to have a prominent link to the eval results for just that model.

I added a link to the evaluations page that sets the `compare` query parameter to only include the relevant fine-tuned model's slug.

> * "Settings" should be its own tab, not a drawer that's always present. Right now if I'm on the "Evaluate" tab and click "Settings" I assume I'm looking at eval settings and get confused.

Agreed, made that change.

> * On "Evaluate" horizontal scrolling should just scroll the table, not the tabs or header

As we discussed, it's more important to 1) show the horizontal scroll bar at all times and 2) be able to scroll the tabs and header off screen.

> * Rename "Output Columns" to "Show Models". Also, this button should be right-justified for consistency with our other pages.

Since we can scroll the button off-screen, I think left-justifying it will provide a smoother experience.

> * There's a weird border radius in the top-left corner of each cell in the header. Should just be straight to the top.

I removed border radius from the column headers.

corbt (approved):
Looks good.

arcticfly:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **TRPCError: No fine tunes found for the given datasetId** `GET /api/trpc/datasets.get,fineTunes.listForDat...` [View Issue](https://openpipe.sentry.io/issues/4552457862/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

Fixed with this PR: https://github.com/OpenPipe/OpenPipe/pull/294

PR #293, opened 2023-10-17
Better type safety for inputs and db operations

corbt:
All the casting `as unknown as ...` makes it really easy to lose track of what types are in each column and received by each endpoint. This PR adds new Zod validators for the JSON columns we work with most frequently and tries to apply them whenever we read from the database so we get good types, and will get alerted if a wrongly-typed value makes it into the database in prod.

Also updates our `/chat/completions` endpoint to be compatible with OpenAI.


app/src/components/fineTunes/ContentTabs/TrainingData/TrainingDataRow.tsx
```diff
@@ -2,7 +2,7 @@ import { useState } from "react";
 import { Th, Td, Thead, Tr, Text, Collapse } from "@chakra-ui/react";
 
 import { type RouterOutputs } from "~/utils/api";
-import { validatedChatInput } from "~/modelProviders/fine-tuned/utils";
+import { chatCompletionInput } from "~/types/shared.types";
```
arcticfly: Is this used?

arcticfly (approved):

PR #294, opened 2023-10-17
Avoid sending error when there are no fine-tunes for dataset

arcticfly:
Previously we were throwing an error when querying fine-tunes for a dataset without any, but instead we should just return an empty array.

Fixes https://openpipe.sentry.io/issues/4552457862/?project=4505642011394048


corbt (commented):
Did you check this works? Theoretically it should complain that we haven't run ACL.

arcticfly:
> Did you check this works? Theoretically it should complain that we haven't run ACL.

Okay _now_ we shouldn't get any more unnecessary errors: https://github.com/OpenPipe/OpenPipe/pull/295

PR #295, opened 2023-10-17
Fix pruning rules

arcticfly:
This PR updates pruning rules to attempt to modify the `messages` field on a dataset entry, instead of the `input` field. It also adds ACL on the `listForDataset` route when no fine-tunes are found.


app/src/server/api/routers/fineTunes.router.ts
```diff
@@ -82,7 +86,10 @@ export const fineTunesRouter = createTRPCRouter({
         .orderBy("ft.createdAt", "desc")
         .execute();
 
-      if (!fineTunes || fineTunes.length === 0) return [];
+      if (!fineTunes || fineTunes.length === 0) {
```
corbt: Technically this is leaking information since you can discover whether someone else's dataset has any models on it. I probably would have queried to get the dataset project id and tested that. But the leakage seems pretty minor in this case.

corbt (approved):

app/src/server/api/routers/fineTunes.router.ts
arcticfly: Good point. We're now running an access control check before returning anything.

PR #297, opened 2023-10-17
Add gpt-3.5-turbo comparison

arcticfly:
Allow user to compare the performance of their models against gpt-3.5-turbo.

### Changes
* Index FineTuneTestingDatasetEntry by `modelId` instead of `fineTuneId`
* In general, allow comparison model id to be used in the same way `fineTuneId` was being used
* Kick off testing jobs when new dataset entries are added to a dataset

### DIFF
--- a/app/scripts/copy-db.sh
+++ b/app/scripts/copy-db.sh
@@ -19,7 +19,7 @@ parse_connection_string() {
 
 # Function to check whether to dump the prod DB
 should_dump_prod_db() {
-    if [ ! -f "/tmp/openpipe-prod.db" ] || [ "$FORCE_DUMP" = "true" ]; then
+    if [ ! -f "$PROD_DUMP_FILE" ] || [ "$FORCE_DUMP" = "true" ]; then
         return 0  # Should dump the prod DB
     else
         return 1  # Should not dump the prod DB
@@ -33,7 +33,7 @@ dump_prod_db() {
     # Set the password as an environment variable
     export PGPASSWORD="$PASSWORD"
     
-    pg_dump -v -Fc -h "$HOST" -U "$DB_USERNAME" -d "$DB_NAME" > /tmp/openpipe-prod.dump
+    pg_dump -v -Fc -h "$HOST" -U "$DB_USERNAME" -d "$DB_NAME" > "$PROD_DUMP_FILE"
     
     # Unset the password environment variable
     unset PGPASSWORD
@@ -60,7 +60,7 @@ create_dev_db() {
 # Function to restore the dump to the dev database
 restore_to_dev_db() {
     report_progress "Restoring dump to dev database..."
-    pg_restore -v --no-owner --no-privileges -d openpipe-dev /tmp/openpipe-prod.dump
+    pg_restore -v --no-owner --no-privileges -d openpipe-dev "$PROD_DUMP_FILE"
 }
 
 # Main execution
--- a/app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
@@ -18,7 +18,12 @@ import { type ComparisonModel } from "@prisma/client";
 import Link from "next/link";
 
 import { api } from "~/utils/api";
-import { useDataset, useHandledAsyncCallback, useSelectedProject } from "~/utils/hooks";
+import {
+  useDataset,
+  useHandledAsyncCallback,
+  useSelectedProject,
+  useTestingEntries,
+} from "~/utils/hooks";
 import { maybeReportError } from "~/utils/errorHandling/maybeReportError";
 import { getComparisonModelName } from "~/utils/baseModels";
 
@@ -39,6 +44,7 @@ const AddComparisonModelDialog = ({
 
   const mutation = api.datasets.update.useMutation();
   const utils = api.useContext();
+  const entries = useTestingEntries().data;
 
   const [onUpdateConfirm, updateInProgress] = useHandledAsyncCallback(async () => {
     if (!dataset?.id || !modelId) return;
@@ -90,7 +96,10 @@ const AddComparisonModelDialog = ({
                 </Text>
               ) : (
                 <VStack>
-                  <Text>To confirm this change, please type the model's ID below.</Text>
+                  <Text>
+                    To confirm this change and run <b>{modelName}</b> against{" "}
+                    <b>{entries?.count || ""}</b> test entries, please type the model's ID below.
+                  </Text>
                   <Box bgColor="orange.100" w="full" p={2} borderRadius={4}>
                     <Text fontFamily="inconsolata">{modelName}</Text>
                   </Box>
--- a/app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
@@ -1,0 +1,128 @@
+import { useRef, useState, useEffect } from "react";
+import {
+  type UseDisclosureReturn,
+  AlertDialog,
+  AlertDialogOverlay,
+  AlertDialogContent,
+  AlertDialogHeader,
+  AlertDialogBody,
+  AlertDialogFooter,
+  Button,
+  VStack,
+  Text,
+  Box,
+  Input,
+  Link as ChakraLink,
+} from "@chakra-ui/react";
+import { type ComparisonModel } from "@prisma/client";
+import Link from "next/link";
+
+import { api } from "~/utils/api";
+import { useDataset, useHandledAsyncCallback, useSelectedProject } from "~/utils/hooks";
+import { maybeReportError } from "~/utils/errorHandling/maybeReportError";
+import { getComparisonModelName } from "~/utils/baseModels";
+
+const AddComparisonModelDialog = ({
+  modelId,
+  disclosure,
+  onClose,
+}: {
+  modelId: ComparisonModel | null;
+  disclosure: UseDisclosureReturn;
+  onClose: () => void;
+}) => {
+  const dataset = useDataset().data;
+  const cancelRef = useRef<HTMLButtonElement>(null);
+
+  const selectedProject = useSelectedProject().data;
+  const needsMissingOpenaiKey = !selectedProject?.condensedOpenAIKey;
+
+  const mutation = api.datasets.update.useMutation();
+  const utils = api.useContext();
+
+  const [onUpdateConfirm, updateInProgress] = useHandledAsyncCallback(async () => {
+    if (!dataset?.id || !modelId) return;
+    const resp = await mutation.mutateAsync({
+      id: dataset.id,
+      updates: { enabledComparisonModels: [...dataset.enabledComparisonModels, modelId] },
+    });
+    if (maybeReportError(resp)) return;
+    await utils.datasetEntries.listTestingEntries.invalidate({ datasetId: dataset.id });
+    await utils.datasets.get.invalidate();
+
+    disclosure.onClose();
+  }, [mutation, dataset?.id, dataset?.enabledComparisonModels, modelId, disclosure.onClose]);
+
+  const [modelNameToConfirm, setModelNameToConfirm] = useState("");
+
+  useEffect(() => {
+    if (disclosure.isOpen) {
+      setModelNameToConfirm("");
+    }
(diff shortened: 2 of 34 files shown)


corbt:
Looks like maybe we aren't passing through function calls properly?
 
Screen Shot 2023-10-17 at 9 22 56 AM

corbt:
Btw I just had a call with a potential client who said they'd much rather use our API key and get a bill than use their own, over rate-limit concerns. Something to keep in mind I guess.

app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
```diff
@@ -0,0 +1,128 @@
...
+                <Text>
+                  To include this model, add your OpenAI API key on the{" "}
+                  <ChakraLink as={Link} href="/project/settings" target="_blank" color="blue.600">
+                    <Text as="span">project settings</Text>
+                  </ChakraLink>{" "}
+                  page.
+                </Text>
+              ) : (
+                <VStack>
+                  <Text>To confirm this change, please type the model's ID below.</Text>
```
corbt: Let's add text like "This will result in 123 calls to the OpenAI API". Right now the dialog does a good job of making this scary but not a great job of giving me the information I need to assess whether I should do it or not.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -20,20 +20,21 @@ export async function getCompletion2(
...
   } else {
     return getModalCompletion(fineTune, prunedInput);
   }
 }
 
-async function getOpenaiCompletion(
-  fineTune: FineTune,
+export async function getOpenaiCompletion(
+  projectId: string,
+  modelId: string | null,
```
corbt: Can't we just get this from the `input` parameter?

app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -20,20 +20,21 @@ export async function getCompletion2(
...
-    return getOpenaiCompletion(fineTune, prunedInput);
+    return getOpenaiCompletion(fineTune.projectId, fineTune.openaiModelId, prunedInput);
   } else {
     return getModalCompletion(fineTune, prunedInput);
   }
 }
 
-async function getOpenaiCompletion(
-  fineTune: FineTune,
+export async function getOpenaiCompletion(
```
corbt: Since this function is no longer fine-tune specific I think it should be moved to another file (probably just included in our openai.ts). It's confusing to have it inside fine-tunes if it's just a generic function to get a completion with a given project's API key.

app/src/server/tasks/evaluateTestSetEntry.task.ts
```diff
@@ -9,48 +10,58 @@ import {
...
-      where: { fineTuneId_datasetEntryId: { fineTuneId, datasetEntryId } },
+      where: { modelId_datasetEntryId: { modelId, datasetEntryId } },
     });
 
     if (existingTestEntry?.output && !skipCache) return;
 
     let prunedMessages = existingTestEntry?.prunedInput;
 
-    const stringsToPrune = await getStringsToPrune(fineTune.id);
+    const stringsToPrune = await getStringsToPrune(modelId);
```
corbt, line 53: Maybe we can just drop pruned messages from the test set entries? Not sure we actually need them there?

app/src/server/tasks/evaluateTestSetEntry.task.ts
```diff
@@ -60,7 +71,7 @@ export const evaluateTestSetEntry = defineTask<EvaluateTestSetEntryJob>({
     }
 
     const cacheKey = hashObject({
-      fineTuneId,
+      modelId,
       input: prunedMessages,
```
corbt: Unrelated to this PR but we need to add the function_call and functions to this hash as well.

app/src/server/tasks/evaluateTestSetEntry.task.ts
```diff
@@ -9,48 +10,58 @@ import {
...
 
-    const datasetEntry = typedDatasetEntry(rawDatasetEntry);
+    let fineTune;
+    if (!isComparisonModel(modelId)) {
+      fineTune = await prisma.fineTune.findUnique({
+        where: { id: modelId },
+      });
+    }
+
+    const datasetEntry = typedDatasetEntry(omit(rawDatasetEntry, "dataset"));
```
corbt: What does this `omit` do?

app/src/server/tasks/evaluateTestSetEntry.task.ts
arcticfly: We definitely do need to prune messages for the test set entries, because some fine-tunes have been trained on pruned messages.

arcticfly:
> Looks like maybe we aren't passing through function calls properly?
> 
> Screen Shot 2023-10-17 at 9 22 56 AM

I'm seeing function call output show up properly. Could you test on another dataset?

app/src/utils/baseModels.ts
```diff
@@ -14,3 +14,13 @@ export const displayBaseModel = (baseModel: BaseModel) => {
...
   }
 };
+
+export const isComparisonModel = (modelId: string) =>
+  ComparisonModel[modelId as keyof typeof ComparisonModel] !== undefined;
+
+export const getComparisonModelName = (comparisonModel: ComparisonModel) => {
+  switch (comparisonModel) {
+    case "GPT_3_5_TURBO":
+      return "gpt-3.5-turbo";
```
corbt: let's add the `-0613` here so we track model changes explicitly.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
arcticfly: No, because sometimes we need to swap out the model that's passed in the input with another model, and reconstruct the input. We could do this swap everywhere we call getOpenaiCompletion, but I think it's cleaner to do it within the function itself.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -82,10 +87,15 @@ export const datasetsRouter = createTRPCRouter({
...
         where: { id: input.id },
         data: {
-          name: input.name,
+          name: input.updates.name,
+          enabledComparisonModels: input.updates.enabledComparisonModels,
         },
       });
 
+      if (input.updates.enabledComparisonModels) {
+        await startDatasetTestJobs(input.id);
```
corbt, line 96: Can we run this in a background job? It took a surprisingly long time for the popup to clear for me.

arcticfly: Agreed, I've noticed the same thing.

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -352,8 +358,8 @@ export const datasetEntriesRouter = createTRPCRouter({
       const newEntry = await prisma.datasetEntry.create({
         data: {
           messages: parsedInput ?? prevEntry.messages,
-          functions: parsedInput ?? prevEntry.functions,
-          function_call: parsedInput ?? prevEntry.function_call,
+          functions: prevEntry.functions ?? undefined,
```
corbt: ouch good catch

corbt (changes requested):
Excited to get this in!

app/src/server/tasks/evaluateTestSetEntry.task.ts
corbt: We need them when we run inference yes. But `getCompletion2` handles that internally. It looks like we aren't using any of `prunedInput`, `prunedInputTokens`, `outputTokens` on the `FineTuneTestingEntry` and could just drop them?

app/src/server/api/routers/datasets.router.ts
corbt: Yeah just tried again with a 10k row dataset and we're 2 minutes in and the button is still spinning. Definitely feels broken.

app/src/server/api/routers/datasetEntries.router.ts
arcticfly: Yeah. I'm not sure how many users were impacted by this. I ended up merging this change in a PR last night: https://github.com/OpenPipe/OpenPipe/pull/296

app/src/modelProviders/fine-tuned/getCompletion-2.ts
corbt: Let's just make the caller swap the model out when constructing the `input` object in those cases. It's really confusing to have an `input.model` parameter which is required but always ignored.

app/src/server/utils/startTestJobs.ts
```diff
@@ -37,6 +68,6 @@ export const startTestJobs = async (fineTune: FineTune) => {
     skipDuplicates: true,
   });
   for (const entry of datasetEntries) {
-    await evaluateTestSetEntry.enqueue({ fineTuneId: fineTune.id, datasetEntryId: entry.id });
```
corbt, line 40: We need to refactor the way this function works so it first figures out which outputs haven't been generated yet and then only does the `createMany`/`enqueue` on the ones that are actually missing. That'll make this way faster and less resource intensive.

It'll also provide a far better UX. I added a gpt-3.5 comparison 10 minutes ago for a dataset with 2500 test entries and 2 fine tuned models and it still hasn't started showing any of the 3.5 completions since it's still working its way through the 5000 no-op jobs for the existing fine-tunes.

app/src/server/tasks/evaluateTestSetEntry.task.ts
arcticfly: Nothing, I removed it. I added it because I was getting a type error and wasn't sure if it was due to the rawDatasetEntry having more fields than typedDatasetEntry expected.

app/src/server/api/routers/datasets.router.ts
arcticfly: We're no longer calculating prunedInputTokens before creating these test entries, which dramatically speeds up the process.

app/src/server/utils/startTestJobs.ts
arcticfly: That's what I'm trying to do on line 35. I suspect a lot of the latency you saw was due to our previous step of calculating input tokens.

app/src/components/datasets/DatasetContentTabs/Evaluation/AddComparisonModelDialog.tsx
arcticfly: Added better feedback:
Screenshot 2023-10-17 at 12 46 44 PM

app/src/server/tasks/evaluateTestSetEntry.task.ts
arcticfly: Added

app/src/server/tasks/evaluateTestSetEntry.task.ts
arcticfly: As we discussed, we'll want to start showing cost soon, so we need prunedInputTokens and outputTokens. I think keeping prunedInput also makes sense (though I suppose we could calculate it on the fly) because if the test output is horribly wrong, it would be useful to see what the pruned input looked like.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
arcticfly: That works for me.

app/scripts/copy-db.sh
```diff
@@ -19,7 +19,7 @@ parse_connection_string() {
 
 # Function to check whether to dump the prod DB
 should_dump_prod_db() {
-    if [ ! -f "/tmp/openpipe-prod.db" ] || [ "$FORCE_DUMP" = "true" ]; then
+    if [ ! -f "$PROD_DUMP_FILE" ] || [ "$FORCE_DUMP" = "true" ]; then
```
corbt, line 22: Where is this variable defined? It should either be a required cli arg or have a default value (or both)

app/src/server/tasks/evaluateTestSetEntry.task.ts
corbt: Yeah let's drop `prunedInput` from the db and calculate it on the fly if necessary.

app/scripts/copy-db.sh
arcticfly: It's defined in your .env file. Since most developers won't use this script, I'm okay with this DX, but you're welcome to improve it.

corbt (approved):

PR #302, opened 2023-10-18
Allow user to sort test dataset entries by model accuracy

arcticfly:
It's important to be able to see what inputs are causing your model to mess up.

### Changes
* Add `useTestEntrySortOrder` hook
* Pass sort order and model slug to `listTestingEntries` route and order by accuracy and sortKey
* Store comparison model slug in url instead of id to determine visibility
* Add helpers to `baseModels` utils file
* Disable auto-refetch on load for logged calls

### DIFF
--- a/app/src/components/datasets/DatasetContentTabs/Evaluation/useTestEntrySortOrder.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/Evaluation/useTestEntrySortOrder.tsx
@@ -1,0 +1,44 @@
+import { useRouter } from "next/router";
+
+import { type SortOrder } from "~/types/shared.types";
+
+const SORT_MODEL_KEY = "sortModel";
+const SORT_ORDER_KEY = "sortOrder";
+
+export const useTestEntrySortOrder = () => {
+  const router = useRouter();
+
+  // Extract the sort model and sort order from the URL query.
+  const sortModelSlug =
+    typeof router.query[SORT_MODEL_KEY] === "string" ? router.query[SORT_MODEL_KEY] : null;
+  const sortOrder =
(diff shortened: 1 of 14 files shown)


arcticfly:
I'm open to feedback on the best/simplest way to optimize the database query to handle cases when the model id is missing or sort order isn't set - right now we always run the left join subquery whether a valid sort model id is provided or not, which seems to work fine in practice, and makes the code simpler.

corbt:
These icons feel too in-your-face. Let's use triangles and gray them out when not activated so they fade into the page more. This kind of pattern: https://datatables.net/examples/basic_init/table_sorting.html

app/src/components/datasets/DatasetContentTabs/Evaluation/useTestEntrySortOrder.tsx
```diff
@@ -0,0 +1,44 @@
+import { useRouter } from "next/router";
+
+import { type SortOrder } from "~/types/shared.types";
+
+const SORT_MODEL_KEY = "sortModel";
+const SORT_ORDER_KEY = "sortOrder";
+
+export const useTestEntrySortOrder = () => {
```
corbt, line 8: We definitely need a general-purpose `useUrlState` hook that persists state as a query param. If `use-query-params` isn't adequate we should write our own; it's silly to keep rewriting this logic.

corbt: To be clear this comment isn't a blocker for this PR, but I think it's something we should do next time we have to rewrite this code, again.

corbt (approved):
Other than the arrow colors/shapes I think this is good.

PR #308, opened 2023-10-19
Allow SDK to skip reporting to OpenPipe

arcticfly:
Add new optional `logRequest` (TypeScript) and `log_request` (Python) fields to SDKs, which default to true. If set to false, the SDK will no longer report to OpenPipe.


corbt (approved):
Thanks looks good.

PR #313, opened 2023-10-20
Save fine tune id on fineTuneTestingEntry

arcticfly:
Previously, we weren't properly saving the fineTuneId on fine tune testing entries. This PR ensures that we do, and removes the step of creating many fine tune testing entries when a new job starts.


corbt:
Did you check this fixes the problem where accuracy wasn't showing up on the list of fine-tuned models? Also, what's the plan to backfill?

PR #316, opened 2023-10-21
Drop support for pipeline version 0

corbt:
Lets us get rid of `getCompletion` and simplify a fair amount of code surrounding it.


app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -23,14 +28,10 @@ export async function getCompletion2(
 
     if (!model) throw new Error("No OpenAI model ID found");
 
-    const completion = await getOpenaiCompletion(fineTune.projectId, {
+    return getOpenaiCompletion(fineTune.projectId, {
       ...prunedInput,
       model,
     });
-    return {
```
corbt, line 30: It was unclear to me why we're override the model on the way out here. Shouldn't it always be the same for fine tunes, and if it weren't wouldn't we want the one OpenAI provides?

app/src/server/tasks/evaluateTestSetEntry.task.ts
```diff
@@ -99,14 +95,14 @@ export const evaluateTestSetEntry = defineTask<EvaluateTestSetEntryJob>({
       model: fineTune
         ? `openpipe:${fineTune.slug}`
         : COMPARISON_MODEL_NAMES[modelId as ComparisonModel],
-      messages: prunedMessages,
```
corbt, line 102: For fine tunes we do the pruning within `getCompletion2` anyway so this is redundant. And for comparison models, we don't actually want the strings to be pruned, so this is unnecessary.

app/src/server/tasks/evaluateTestSetEntry.task.ts
```diff
@@ -99,14 +95,14 @@ export const evaluateTestSetEntry = defineTask<EvaluateTestSetEntryJob>({
...
+      messages: datasetEntry.messages,
       function_call: datasetEntry.function_call ?? undefined,
       functions: datasetEntry.functions ?? undefined,
     };
     try {
       if (isComparisonModel(modelId)) {
         completion = await getOpenaiCompletion(rawDatasetEntry.dataset.projectId, input);
-      } else if (fineTune && (fineTune.pipelineVersion === 1 || fineTune.pipelineVersion === 2)) {
```
corbt, line 109: This check happens within `getCompletion2`

PR #323, opened 2023-10-27
Add filters to General and Evaluate tabs

arcticfly:
Add input and output filters to dataset entries in the General and Evaluate tabs. We still need to add filters to dataset entry type in the General tab, and dataset entry provenance in the Evaluate tab.

### Changes
* Use enums to standardize filter field names between client and server
* Add filters to `General` and `Evaluate` tabs
* Move `deployedFineTunes` field from `datasetEntries.testingStats` to `datasets.get`
* Color-code TRAIN and TEST rows

### DIFF
--- a/app/src/types/shared.types.ts
+++ b/app/src/types/shared.types.ts
@@ -92,3 +92,13 @@ export enum SortOrder {
   ASC = "asc",
   DESC = "desc",
 }
+
+export const comparators = ["=", "!=", "CONTAINS", "NOT_CONTAINS"] as const;
+
+export const filtersSchema = z.array(
+  z.object({
+    field: z.string(),
+    comparator: z.enum(comparators),
+    value: z.string(),
+  }),
+);
--- a/app/src/types/shared.types.ts
+++ b/app/src/types/shared.types.ts
@@ -102,3 +102,20 @@ export const filtersSchema = z.array(
     value: z.string(),
   }),
 );
+
+export enum LoggedCallsFiltersDefaultFields {
+  Request = "Request",
+  Response = "Response",
+  Model = "Model",
+  StatusCode = "Status Code",
+}
+
+export enum GeneralFiltersDefaultFields {
+  Input = "Input",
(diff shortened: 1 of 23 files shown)


app/src/types/shared.types.ts
```diff
@@ -92,3 +92,32 @@ export enum SortOrder {
...
+
+export const filtersSchema = z.array(
+  z.object({
+    field: z.string(),
+    comparator: z.enum(comparators),
+    value: z.string(),
+  }),
+);
+
+export enum LoggedCallsFiltersDefaultFields {
```
corbt, line 106: There's almost never a good reason to use a Typescript `enum` (in fact I believe I heard an interview with the TS architect who mentioned he regrets putting them in the language).

Union types are generally a better way to do the same thing. In this case I'd rewrite it as something like:

```ts
export const loggedCallsFiltersDefaultFields = ["Request", "Response", "Model", "Status Code"] as const;

export const LoggedCallsFiltersFieldTypes = (typeof loggedCallsFiltersDefaultFields)[number]
```

corbt (approved):
Exciting, already tested locally and helps me get a better sense of model perf on the evaluate tab!

corbt:
Oh one change that i think is pretty important: if any filters are applied in the url we should show the filter box on page load by default. It's confusing to load a page with filters applied but the filter box is closed.

app/src/types/shared.types.ts
arcticfly: In this case, I think an Enum is the best way to compare on a per-field basis when constructing the filters query, e.g.

```
if (filter.field === EvaluationFiltersDefaultFields.Input) {
  wheres.push(filterExpression(sql.raw(`de."messages"::text`)));
}
```

arcticfly:
> Oh one change that i think is pretty important: if any filters are applied in the url we should show the filter box on page load by default. It's confusing to load a page with filters applied but the filter box is closed.

Good thinking, I added that change.

PR #327, opened 2023-10-30
Add date filter to request logs

arcticfly:
With this change, we can now filter request logs by date.

### Changes
* Add `text | date` type field to filter option
* Split filter into TextFilter and DateFilter
* Add `LAST 15M`, `LAST 24H`, `LAST 7D`, `BEFORE`, `AFTER` and `RANGE` date comparators
* Add `"Sent At"` as a filterable field for request logs

Sent At filters:
Screenshot 2023-10-30 at 1 15 13 AM

### DIFF
--- a/app/src/components/Filters/AddFilterButton.tsx
+++ b/app/src/components/Filters/AddFilterButton.tsx
@@ -1,10 +1,10 @@
 import { Button, HStack, Icon, Text } from "@chakra-ui/react";
 import { BsPlus } from "react-icons/bs";
 import { useFilters } from "./useFilters";
-import { comparators } from "~/types/shared.types";
+import { type AtLeastOne, comparators } from "~/types/shared.types";
 import { type FilterOptionType } from "./types";
 
-const AddFilterButton = ({ filterOptions }: { filterOptions: FilterOptionType[] }) => {
+const AddFilterButton = ({ filterOptions }: { filterOptions: AtLeastOne<FilterOptionType> }) => {
   const addFilter = useFilters().addFilter;
 
   return (
@@ -14,7 +14,7 @@ const AddFilterButton = ({ filterOptions }: { filterOptions: FilterOptionType[] 
       onClick={() =>
         addFilter({
           id: Date.now().toString(),
-          field: filterOptions[0]?.field as string,
+          field: filterOptions[0].field,
           comparator: comparators[0],
           value: "",
         })
--- a/app/src/components/Filters/AddFilterButton.tsx
+++ b/app/src/components/Filters/AddFilterButton.tsx
@@ -2,8 +2,9 @@ import { Button, HStack, Icon, Text } from "@chakra-ui/react";
 import { BsPlus } from "react-icons/bs";
 import { useFilters } from "./useFilters";
 import { comparators } from "~/types/shared.types";
+import { type FilterOptionType } from "./types";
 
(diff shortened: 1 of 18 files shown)


app/src/components/Filters/AddFilterButton.tsx
```diff
@@ -1,8 +1,10 @@
 import { Button, HStack, Icon, Text } from "@chakra-ui/react";
 import { BsPlus } from "react-icons/bs";
-import { comparators, useFilters } from "./useFilters";
+import { useFilters } from "./useFilters";
+import { comparators } from "~/types/shared.types";
+import { type FilterOptionType } from "./types";
 
-const AddFilterButton = ({ filterOptions }: { filterOptions: string[] }) => {
+const AddFilterButton = ({ filterOptions }: { filterOptions: FilterOptionType[] }) => {
```
corbt: Minor, but we could define a helper type like follows:

`type AtLeastOne<T> = [T, ...T[]];`

And then change the type here to `filterOptions: AtLeastOne<FilterOptionType>`.

That way TS would enforce there's at least one filter option passes, so you wouldn't have to cast the `as string` below.

app/src/components/Filters/Filter/DateFilter.tsx
```diff
@@ -0,0 +1,102 @@
...
+      updateDate("");
+    }
+  }, [valueIsEmpty, updateDate]);
+
+  return (
+    <HStack>
+      <SelectFieldDropdown filterOptions={filterOptions} filter={filter} />
+      <SelectComparatorDropdown filter={filter} filterType="date" />
+      <HStack>
+        {!["LAST 15M", "LAST 24H", "LAST 7D"].includes(filter.comparator) && (
```
corbt, line 71: This is a nit but can we update the way we display these at least to be title case (eg. "Last 15M", "Before")? They kinda feel like they're shouting at me. 😁

app/src/server/utils/constructLoggedCallFiltersQuery.ts
```diff
@@ -31,8 +29,41 @@ const comparatorToSqlExpression = (comparator: (typeof comparators)[number], val
   };
 };
 
-export const constructFiltersQuery = (
-  filters: z.infer<typeof logFiltersSchema>,
+export const dateComparatorToSqlExpression = (
+  comparator: (typeof comparators)[number],
+  value: number[],
```
corbt: It loks like this is always supposed to be of length 2? If so you can define it as a tuple eg. `value: [number, number]`.

app/src/server/utils/constructLoggedCallFiltersQuery.ts
```diff
@@ -46,20 +77,41 @@ export const constructFiltersQuery = (
     .where((eb) => {
       const wheres: Expression<SqlBool>[] = [eb("lc.projectId", "=", projectId)];
 
+      console.log("filters", filters);
```
corbt: rm

corbt (approved):
Few small comments but looks good!

app/src/components/Filters/AddFilterButton.tsx
arcticfly: Yes, this does save some casting (unfortunately it adds an `unknown` casting in `Filters.tsx`, but at least the reasoning behind the cast is explained by its surrounding code).

PR #331, opened 2023-11-01
Relabel with GPT-4

arcticfly:
### Changes
* Add dataset entry provenance
* Add dataset entry history
* Add job to relabel with GPT-4
* Show relabeling status in dataset entries table


app/prisma/schema.txt
```diff
@@ -0,0 +1,697 @@
+// This is your Prisma schema file,
```
corbt: rm

app/prisma/migrations/20231031174505_add_relabel_request/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+-- CreateEnum
+CREATE TYPE "RelabelRequestStatus" AS ENUM ('PENDING', 'IN_PROGRESS', 'COMPLETE', 'ERROR');
+
+-- Add new columns with default values
+ALTER TABLE "DatasetEntry"
+ADD COLUMN "importId" TEXT NOT NULL DEFAULT '',
+ADD COLUMN "provenance" "DatasetEntryProvenance" NOT NULL DEFAULT 'UPLOAD';
+
+-- Update existing rows to use the default values
+UPDATE "DatasetEntry" SET "importId" = '' WHERE "importId" IS NULL;
```
corbt, line 20: Nit but I think this line and the next are actually unnecessary, the `DEFAULT` that you set when creating the column should already take care of this. No need to change it now though.

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/DatasetEntryEditorDrawer.tsx
```diff
@@ -106,67 +111,120 @@ export default function DatasetEntryEditorDrawer({
...
+                  <>
+                    <Text>
+                      <Text as="span" fontWeight="bold">
+                        Import ID:
+                      </Text>{" "}
+                      {datasetEntry.importId}
+                    </Text>
+                  </>
+                )}
+                <Text>
```
corbt, line 148: We might be able to both fit the visual styling of the drawer better and take up less space by removing the outline/different-colored background but not blocking, I can play with that later.

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/DatasetEntriesTable.tsx
```diff
@@ -1,33 +1,55 @@
...
 import { TableHeader, TableRow, EmptyTableRow } from "./TableRow";
 import DatasetEntryEditorDrawer from "./DatasetEntryEditorDrawer";
 
 export default function DatasetEntriesTable() {
   const [expandedDatasetEntryId, setExpandedDatasetEntryId] = useState<string | null>(null);
-  const datasetEntries = useDatasetEntries().data?.entries;
+  const [refetchInterval, setRefetchInterval] = useState(0);
+  const datasetEntries = useDatasetEntries(refetchInterval).data?.entries;
+
+  const relabelingVisible = useMemo(() => {
```
corbt, line 14: Nice touch, the column helps a lot with making it clear what's going on.

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -466,6 +498,46 @@ export const datasetEntriesRouter = createTRPCRouter({
...
+        },
+      });
+
+      if (!dataset) {
+        return error("Dataset not found for dataset entry");
+      }
+
+      await requireCanModifyProject(dataset.projectId, ctx);
+
+      const batchId = Date.now().toString();
```
corbt: Can we use `new Date().toISOString()` here instead? Will be more interpretable when it shows up in the filters.

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -71,6 +72,13 @@ export const datasetEntriesRouter = createTRPCRouter({
             "de.updatedAt as updatedAt",
             "de.outdated as outdated",
             "de.datasetId as datasetId",
+            jsonArrayFrom(
```
corbt, line 75: This seems good enough; probably will only have O(1) relabel requests per entry anyway so not worth optimizing.

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -299,7 +322,14 @@ export const datasetEntriesRouter = createTRPCRouter({
         })
         .filter(truthyFilter);
 
-      const datasetEntriesToCreate = await prepareDatasetEntriesForImport(datasetId, rowsToConvert);
+      const importId = Date.now().toString();
```
corbt: `new Date().toISOString()` for readability

app/src/server/tasks/importDatasetEntries.task.ts
```diff
@@ -79,11 +80,15 @@ export const importDatasetEntries = defineTask<ImportDatasetEntriesJob>({
       },
     });
 
+    const importId = Date.now().toString();
```
corbt: `new Date().toISOString()` for readability.

app/src/server/utils/prepareDatasetEntriesForImport.ts
```diff
@@ -29,7 +32,10 @@ export const prepareDatasetEntriesForImport = async (
   existingTrainingCount += entriesToImport.filter((row) => row.split === "TRAIN").length;
 
   const newTotalEntries = existingCount + entriesToImport.length;
-  const numTrainingToAdd = Math.floor(trainingRatio * newTotalEntries) - existingTrainingCount;
+  const numTrainingToAdd = Math.min(
```
corbt: Curious what drove this change -- I don't immediately see the bug in the code as it was written previously but maybe I'm missing something?

corbt (changes requested):
Minor changes but nice work, excited to have this available!

app/prisma/migrations/20231031174505_add_relabel_request/migration.sql
arcticfly: Yes you're right. I forgot to remove these lines after editing an earlier version.

app/src/server/utils/prepareDatasetEntriesForImport.ts
arcticfly: It fixed a bug in the previous version of the code, but I didn't see how much you had changed. I think the changes you made fixed the bug that this is supposed to fix, so I can remove it now.

app/src/server/utils/prepareDatasetEntriesForImport.ts
```diff
@@ -26,12 +29,10 @@ export const prepareDatasetEntriesForImport = async (
 
   const trainingRatio = dataset?.trainingRatio ?? 0.8;
 
-  existingTrainingCount += entriesToImport.filter((row) => row.split === "TRAIN").length;
-
   const newTotalEntries = existingCount + entriesToImport.length;
   const numTrainingToAdd = Math.floor(trainingRatio * newTotalEntries) - existingTrainingCount;
 
-  let numTrainingAdded = 0;
+  let numTrainingAdded = entriesToImport.filter((row) => row.split === "TRAIN").length;
```
arcticfly, line 35: I think it's more clear to set `numTrainingAdded` to actually encompass the total number of training rows added in this call. Same goes for `existingTrainingCount`, it should be equal to the number of training rows that existed before this route was called.

app/src/server/api/routers/datasetEntries.router.ts
arcticfly: Works for me

app/src/server/utils/prepareDatasetEntriesForImport.ts
corbt: I agree that with your new changes the code is cleaner than the way I wrote it initially.

corbt (approved):
Nice work.

PR #332, opened 2023-11-03
Properly report errors through SDKs

arcticfly:
### Changes
* Report errors in typescript sdk
* Raise errors and report with proper error codes in python sdk
* Increment versions


corbt (approved):
Haven't tested the code but looks reasonable.

PR #336, opened 2023-11-07
Update SDKs and server to match new OpenAI spec

### LINKED ISSUE #335: support openai lib: 4.15.3
the code here only works for `4.8.0`: https://github.com/OpenPipe/OpenPipe/blob/main/client-libs/typescript/src/openai/streaming.ts#L13

in the latest version, current(`4.15.3`): it is here: https://github.com/openai/openai-node/blob/e0138e8a5939f6fe53639df78cd541e4e20927ed/src/streaming.ts#L21C24-L21C24

related PR: https://github.com/openai/openai-node/pull/370
The problem is that current OpenPipe's OpenAI lib does not work for `streaming` API due to refactoring of the PR above. 

can we update to support latest code?

@arcticfly @corbt

arcticfly:
Fixes https://github.com/OpenPipe/OpenPipe/issues/335


app/src/server/modal-rpc/clients.ts
```diff
@@ -26,7 +26,7 @@ type ModalInput = {
   model: string;
   prompt: string;
   n?: number;
-  max_tokens?: number;
+  max_tokens?: number | null;
```
arcticfly: Will Modal handle this properly by default?

corbt: Good catch. No, vLLM throws an error. So inside `runInference` we should check if this key is `null` and remove it from the args before sending the request to Modal.

arcticfly: Probably easier to not pass `max_tokens: null` to this function, and pass undefined instead.

app/src/types/dbColumns.types.ts
```diff
@@ -37,7 +38,7 @@ export const typedLoggedCallModelResponse = <T extends Pick<LoggedCallModelRespo
 
 const fineTuneTestingEntrySchema = z
   .object({
-    output: chatMessage.passthrough().optional(),
+    output: chatCompletionMessage,
```
arcticfly: Does this need to be optional?

corbt: FineTuneTestingEntry.output is nullable in the database, so I think so? I actually don't remember when that field gets set relative to model creation but if we could make it non-nullable in the db then we can definitely make it non-nullable here.

PR #338, opened 2023-11-09
Support tool calls

arcticfly:
This PR allows us to support tool calls while continuing to serve inference with models trained on function calls. However, from the perspective of our dataset entries, function calling will no longer exist. This PR contains a migration file to rewrite all existing dataset entry function metadata to fit the tool calling syntax, and from now on we'll convert all rows from request logs and uploaded files to fit the tool calling syntax as well. Request logs will still show function calls, and requests made using the function call syntax will receive responses in the same syntax.

### Changes
* Add `tool_choice` and `tools` fields to dataset entry schema
* Add script to migrate all references to functions in `function_call` and `functions`, along with `function` and `assistant` messages, to new tool calling syntax
* Rewrite our scoring function
* Update the dataset entry editor to be able to add and remove tool calls within a message
* Fix various bugs within the editor
* Update serialization/deserialization code to expect tool calling syntax
* Conditionally reconvert inference output from tool calling to function call if function call syntax was used in the request

Tool call editor:
Screenshot 2023-11-09 at 12 47 03 AM


app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/parseableToToolCalls.ts
```diff
@@ -0,0 +1,70 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let parsedJSON: any;
+  try {
+    parsedJSON = JSON.parse(str);
+  } catch {
+    return false;
+  }
+
+  // Check if the parsedJSON is an object and not null
+  if (parsedJSON === null || !Array.isArray(parsedJSON)) {
```
corbt: I think this function would be a lot shorter, cleaner and more maintainable if it were just two zod validators, one for the array of objects, and then a second one for the arguments?

app/src/modelProviders/fine-tuned/serializers.ts
```diff
@@ -5,11 +5,13 @@ const FUNCTION_CALL_TAG = "<function>";
 const FUNCTION_ARGS_TAG = "<arguments>";
```
corbt, line 10: This file makes sense but I'd feel more comfortable if we had a couple of tests here with both tool-call and function inputs for sanity.

app/src/server/utils/convertFunctionCalls.ts
```diff
@@ -0,0 +1,96 @@
+import type { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
+
+export const convertFunctionCall = (
+  functionCall: ChatCompletionCreateParamsBase["function_call"] | null,
+): ChatCompletionCreateParamsBase["tool_choice"] => {
+  if (!functionCall) {
+    return undefined;
```
corbt, line 7: This seems a little concerning to me, in the sense that it could take a null and return an undefined, which could lead to unexpected errors. I'd expect it to return the same thing it got if it isn't a valid function call, or at least to be explicit about the conversion in the type signature and/or comments.

app/src/server/utils/convertFunctionCalls.ts
```diff
@@ -0,0 +1,96 @@
...
+      description: func.description,
+      parameters: func.parameters,
+    },
+  }));
+};
+
+export const convertMessages = (
+  messages: ChatCompletionCreateParamsBase["messages"],
+): ChatCompletionCreateParamsBase["messages"] =>
+  messages ? messages.map(convertFunctionMessageToToolCall) : [];
```
corbt: Can messages ever be null? If not, this check isn't necessary. And if it can be, we should just return the null, not coerce it to an empty array.

app/src/server/utils/convertFunctionCalls.ts
```diff
@@ -0,0 +1,96 @@
...
+          tool_calls: [
+            {
+              id: "",
+              type: "function" as const,
+              function: message.function_call,
+            },
+          ],
+        };
+      }
+    case "function":
```
corbt, line 61: I clearly need to read the new docs, but what's the difference between a `role: "assistant"` message that includes "tool_calls" field and a `role: "tool"`/`role: "function"` message?

app/src/server/utils/prepareDatasetEntriesForImport.ts
```diff
@@ -54,10 +60,15 @@ export const prepareDatasetEntriesForImport = async (
     return {
       id: uuidv4(),
       datasetId: datasetId,
-      messages: row.input.messages as object[],
+      messages: convertMessages(row.input.messages) as object[],
       function_call: row.input.function_call as object,
       functions: row.input.functions as object[],
-      output: (row.output as unknown as Prisma.InputJsonValue) ?? {
+      tool_choice:
+        row.input.tool_choice || (convertFunctionCall(row.input.function_call) as object),
```
corbt, line 67: I don't like this casting; why aren't we getting the right types from `convertFunctionCall`? Are we sure this is always an object? Couldn't a message have neither `tool_choice` or `function_call`, in which case it would be undefined or null? We have control over the output of this function; we should make sure we do the correct typechecks if necessary and that the types are right without having to cast.

Same thing on lines 63 and 68.

corbt (changes requested):
Nice work. I think the schema changes you made and the places you convert to/from tool_calls are exactly right, which will minimize the headache we have to deal with while we go through the transition.

I do have some small comments, the most important ones being around type safety -- want to make sure we know what types we're dealing with everywhere as the code handling this gets more complex.

app/src/server/utils/prepareDatasetEntriesForImport.ts
arcticfly: It's not always an object. It could be undefined or a string as well. But we need to cast it to something that prisma is willing to save in the tool_choice field, same as we were doing with function_call and functions. If it's actually null and we cast it to an object, that doesn't affect the saving process.

app/src/server/utils/convertFunctionCalls.ts
arcticfly: https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/parseableToToolCalls.ts
arcticfly: Ah yes, I had forgotten how wonderful zod is.

app/src/modelProviders/fine-tuned/serializers.ts
arcticfly: Good idea, I caught a bug while testing.

app/src/server/utils/convertFunctionCalls.ts
arcticfly: My intent was to allow all callers to rely on `convertFunctions` returning a value that matches the `tool_choice` type, but I see how that could be confusing. I'll modify the type signature and return null as well.

app/src/server/utils/convertFunctionCalls.ts
arcticfly: No they can't. This was left over from an earlier refactor. Good catch.

corbt (approved):

PR #347, opened 2023-11-13
Convert tool calls to function calls for 3.5 fine-tunes

arcticfly:
Unfortunately, OpenAI's 3.5 fine-tuning services have not been updated to support the tool call syntax. Consequently, we need to convert incoming requests to fine-tuned 3.5 models to use the function call syntax, and then convert completions generated from fine-tuned 3.5 models back into the tool call syntax to ensure consistency with the remainder of our app.

### DIFF
--- a/app/src/server/utils/convertFunctionCalls.ts
+++ b/app/src/server/utils/convertFunctionCalls.ts
@@ -1,11 +1,9 @@
 import type { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
 
-export const convertFunctionCall = (
+export const convertFunctionCallToToolChoice = (
   functionCall: ChatCompletionCreateParamsBase["function_call"] | null,
-): ChatCompletionCreateParamsBase["tool_choice"] => {
-  if (!functionCall) {
-    return undefined;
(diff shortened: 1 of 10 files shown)


app/src/server/utils/convertFunctionCalls.ts
```diff
@@ -31,7 +41,29 @@ export const convertFunctions = (
   }));
 };
 
-export const convertMessages = (
+export const convertToolsToFunctions = (
+  tools: ChatCompletionCreateParamsBase["tools"],
+): ChatCompletionCreateParamsBase["functions"] => {
+  if (!tools) return tools;
+  return tools.map((tool) => ({
```
corbt: I'm probably missing something but just looking at the logic couldn't this be written as `return tools?.map((tool) => tool.function)`?

corbt (approved):
I can't think of a better way to do this; looks good. Thanks for getting it in!

app/src/server/utils/convertFunctionCalls.ts
arcticfly: No, I think you're right. I had it structured this way because of some legacy logic that I've since removed, so I'll update it to be sleeker.

PR #348, opened 2023-11-13
Upgrade python version

arcticfly:
The OpenAI python spec has noticeably changed between versions `0.27.8` and `1.2.3`, so the refactor to get up to date is significant. Notably, users must now create a client object before making requests to OpenAI servers, similarly to the TypeScript SDK.


corbt (approved):
Ok yeah let's stop using the deprecated fields but yes, this looks reasonable!

PR #355, opened 2023-11-17
Pass verify_ssl to configure_openpipe_client

DmitryBe:
Disable ssl verification
```
client = OpenAI(openpipe={"verify_ssl": False})
```


arcticfly (approved):
Makes sense, thanks for adding this.

PR #357, opened 2023-11-18
Polish instructions to run OpenPipe locally

bjsi:
The setup process is already quite smooth. Just adding some extra detail based on my experience getting it up and running!


arcticfly (approved):
Looks great, thanks for adding this!

PR #358, opened 2023-11-19
Bump deps

corbt:
Apparently there's a vulnerability in Babel this fixes. Unlikely to actually affect us but bumping just in case.


arcticfly (approved):
Oh no, do we have to review everything now?

I guess that gives me job security.

PR #359, opened 2023-11-19
Fix crashes caused by isolated VM package

bjsi:
Pass --no-node-snapshot in NODE_OPTIONS to the `dev:next`, `dev:worker` and `seed` commands. This is required according to the maintainers of isolated VM https://github.com/laverdet/isolated-vm/issues/420. This was causing difficult to debug issues where any trpc requests would return a 500 internal server error with this unhelpful error message. I was able to track it down to this cause by commenting out sections of server code until I found it.

```sh
[next] - error Error: socket hang up
[next]     at connResetException (node:internal/errors:721:14)
[next]     at Socket.socketOnEnd (node:_http_client:519:23)
[next]     at Socket.emit (node:events:526:35)
[next]     at Socket.emit (node:domain:488:12)
[next]     at endReadableNT (node:internal/streams/readable:1408:12)
[next]     at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
[next]   code: 'ECONNRESET'
[next] }
```

Not sure if it's because I'm running on an M2 Mac? But adding the node option fixes it.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build": "next build",
-    "dev:next": "TZ=UTC next dev",
+    "dev:next": "NODE_OPTIONS='--no-node-snapshot' TZ=UTC next dev",
     "dev:wss": "tsx --watch src/wss-server.ts",
     "dev:worker": "tsx --watch src/server/tasks/worker.ts",
     "worker": "tsx src/server/tasks/worker.ts",
--- a/app/package.json
+++ b/app/package.json
@@ -12,7 +12,7 @@
     "build": "next build",
     "dev:next": "NODE_OPTIONS='--no-node-snapshot' TZ=UTC next dev",
     "dev:wss": "tsx --watch src/wss-server.ts",
-    "dev:worker": "tsx --watch src/server/tasks/worker.ts",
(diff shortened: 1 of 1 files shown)


corbt:
Oh interesting. Were you using any of the functionality provided by isolated-vm or these crashes were happening for completely unrelated requests? A better fix honestly might just be to rip it out; it was only used in the experiments view which we've deemphasized and hidden at this point anyway.

bjsi:
> Were you using any of the functionality provided by isolated-vm or these crashes were happening for completely unrelated requests?

Crashes were happening for unrelated requests. I think it's because isolated VM objects get instantiated when exported functions like `deriveNewConstructFn` and `parsePromptConstructor` are imported (even though they aren't actually used in the request). Since some of the trpc routers import and use these functions, all of my requests would fail for seemingly no reason.

>  A better fix honestly might just be to rip it out

Sure, I could handle that 👍

corbt (approved):
Ok let's merge. Unclear to me why this was causing a crash on your hardware and not ours but could be M2 related for sure.

At some point we should probably remove the experiments-related code entirely to simplify the codebase, but don't need to do that right now.

Thanks for this!

PR #362, opened 2023-11-22
Dataset evals

arcticfly:
Add head-to-head comparison evals, and migrate our previous field comparison evals to fit into the new schema.

### Changes
* Add Evals dropdown
* Add eval stats to model headers
* Add individual comparison stats to each evaluation cell
* Add head-to-head comparison modal
* Add field comparison modal


app/src/server/tasks/generateTestSetEntry.task.ts
```diff
@@ -11,20 +12,27 @@ import {
...
-export const evaluateTestSetEntry = defineTask<EvaluateTestSetEntryJob>({
-  id: "evaluateTestSetEntry",
+export const generateTestSetEntry = defineTask<GenerateTestSetEntryJob>({
+  id: "generateTestSetEntry",
   handler: async (task) => {
-    const { modelId, datasetEntryId, skipCache } = task;
+    const { modelId, datasetEntryId } = task;
+
+    // TODO: fix evals with caching
+    const skipCache = true;
```
corbt: Under what circumstances will this cause a re-run of all outputs/evals?

app/src/server/tasks/relabelDatasetEntryTask.ts
```diff
@@ -106,6 +108,8 @@ export const relabelDatasetEntry = defineTask<RelabelDatasetEntryJob>({
       ]);
 
       if (newDatasetEntry.split === "TEST") {
+        await copyDatasetEvalDatasetEntries(datasetEntry.id, newDatasetEntry.id);
+        await updatePruningRuleMatches(datasetEntry.datasetId, new Date(0), [newDatasetEntry.id]);
```
corbt: Don't we want to update pruning rule matches even if the split is TRAIN?

app/src/server/utils/copyDatasetEvalDatasetEntries.ts
```diff
@@ -0,0 +1,19 @@
+import { prisma } from "../db";
+
+export const copyDatasetEvalDatasetEntries = async (
```
corbt, line 3: Will this always be used in the same place we call `updatePruningRules`, which are the same places we're creating updated dataset entries?

I think we should create a helper `updateDatasetEntry` that takes in an entry id as well as an object of fields to update, then does the update and calls these as needed. Too easy to forget a place otherwise, like I did with `output_format` (and I don't think I ever fixed, actually).

app/src/server/api/routers/datasetEvals.router.ts
```diff
@@ -0,0 +1,465 @@
...
+          `The test set only has ${testDatasetEntries.length} entries, but ${input.numDatasetEntries} were requested`,
+        );
+      }
+
+      const shuffledEntryIds = shuffle(testDatasetEntries.map((entry) => entry.id)).slice(
+        0,
+        input.numDatasetEntries,
+      );
+
+      console.log("shuffledEntryIds", shuffledEntryIds);
```
corbt: rm

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+
+    const instructions = firstResult.datasetEvalDatasetEntry.datasetEval.instructions;
+
+    try {
+      const input: ChatCompletionCreateParams = {
+        model: "gpt-4",
+        messages: [
+          {
+            role: "system",
+            content: `You are an intelligent and fair judge of chatbots. Evaluate the following two chatbot responses and choose which one is better in the following way: ${
```
corbt: You are an intelligent and fair judge of chatbots. Evaluate the following two chatbot responses and choose which one is better. If the outputs are of similar quality, you can mark them as `EQUAL`.

And then, if there are instructions, add the following:

The user has provided the following instructions on what you should evaluate the outputs on: <instructions>

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+          },
+          {
+            role: "user",
+            content: `This is what Alice said:\n\n${JSON.stringify(firstEntry.output)}`,
+          },
+          {
+            role: "user",
+            content: `This is what Bob said:\n\n${JSON.stringify(secondEntry.output)}`,
+          },
+          {
```
corbt: I'd remove this final message unless you've tried and it makes a noticeable positive impact on the quality.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+        input,
+      );
+
+      const args = response.choices[0]?.message?.tool_calls?.[0]?.function?.arguments;
+
+      if (!args) throw new Error("No arguments returned" + JSON.stringify(response));
+
+      const parsedArgs = JSON.parse(args);
+
+      if (!parsedArgs["explanation"] || !parsedArgs["judgement"]) {
```
corbt: Easier to just define a zod object for this and use that to validate. You can also use https://www.npmjs.com/package/zod-to-json-schema to use the same zod object as the function definition you pass to openai.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+      }
+
+      explanation = parsedArgs["explanation"] as string;
+      judgement = parsedArgs["judgement"] as (typeof JUDGEMENT_OPTIONS)[number];
+
+      if (!JUDGEMENT_OPTIONS.includes(judgement)) {
+        throw new Error("Invalid judgement returned" + JSON.stringify(response));
+      }
+    } catch (e) {
+      console.error("error getting judgement", e);
```
corbt: Let's report this error to sentry.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+          },
+          {
+            role: "user",
+            content: formatDatasetEntryInputInstructions(
+              firstResult.datasetEvalDatasetEntry.datasetEntry,
+            ),
+          },
+          {
+            role: "user",
+            content: `This is what Alice said:\n\n${JSON.stringify(firstEntry.output)}`,
```
corbt: I'd lean towards just saying "This is response_1" and "This is response_2". Any reason to believe calling them "Alice" and "Bob" would work better? I'd be worried about that introducing more bias in the analysis, not less.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -0,0 +1,428 @@
...
+          {
+            type: "function",
+            function: {
+              name: "record_score",
+              parameters: {
+                type: "object",
+                properties: {
+                  explanation: {
+                    type: "string",
+                    description: "An explanation of why you chose the response you did",
```
corbt: "Explain any factors that might make one response better than the other, and then state which one is higher quality overall. If they're about the same you can just say that."

corbt (approved):
I'll be honest and say I don't have a full mental model of the "execution graph" of when comparisons are added and what happens when we update the models or dataset entries associated with an eval. But the schema seems reasonable and if we run into issues later around perf we can fix those.

A few comments/suggestions (full disclosure, didn't read much of the frontend code at all) but looks good!

app/src/server/tasks/generateTestSetEntry.task.ts
arcticfly: The purpose of this task is to query and save output created by a comparison or fine-tuned model based on input from the dataset entry. The cache is meant to be used in two ways:

1. Prevent this task from running on the same combination of dataset entry and model id as it has in the past
2. Save this task the work of re-retrieving output when presented with identical input used in a previous retrieval run. Instead of retrieving the output a second time from the same model with the same input, we copy over the output that was retrieved the first time we ran the task with this input and modelId.

In scenario 1, skipping the cache could indeed cause unnecessary and duplicate evals to be run. So if we call generateTestSetEntry on an identical dataset entry and model id many times, then we'll duplicate the evals many times as well. This would be very inefficient and costly. However, I don't believe our code is designed to call generateTestSetEntry multiple times with the same input.

In scenario 2, skipping the cache will result in an extra inference request to calculate the output for this model id and dataset entry combination, but will not result in extra eval calls. Since our evals are comparative (and since the explanation as currently designed requires the model's true name to be available while constructing it) we need to re-run these evals every time we generate new output, whether we generate it from a request to an LLM or from cached output.

app/src/server/tasks/relabelDatasetEntryTask.ts
arcticfly: Yes we do, good catch.

app/src/server/utils/copyDatasetEvalDatasetEntries.ts
arcticfly: I think that's a great idea. I'd like to do that in a separate PR, since there's already a lot of refactoring going on in this one.

app/src/server/tasks/evaluateTestSetEntries.task.ts
arcticfly: Let's experiment without it. I included it because I've heard LLMs pay the most attention to the very beginning and ending of inputs, and I want to make sure that these specific instructions are observed. However, it's possible including them in the system message will be sufficient.

app/src/server/tasks/generateTestSetEntry.task.ts
arcticfly: In any case, I fixed our evals to work with caching, so this should be an non-issue.

app/src/server/tasks/evaluateTestSetEntries.task.ts
arcticfly: Good idea, using zod now.

app/src/server/tasks/evaluateTestSetEntries.task.ts
arcticfly: It does seem that adding this at the end is important to keeping the judgement focused on the particular criteria the user specifies.

app/src/server/tasks/evaluateTestSetEntries.task.ts
corbt: @arcticfly any feedback on this?

arcticfly: Personally, it would be easier for me to keep track of Alice's and Bob's answers rather than response_1 and response_2, which is why I leaned toward labeling the outputs that way. However, that advantage may not translate to GPT-4. I'd be totally fine with you experimenting with the wording here, and hopefully it won't require very much context loading.

PR #368, opened 2023-11-26
Add header to original output row

arcticfly:
Add header to original output row. For head to head evals, also show wins, ties, and losses in the eval dropdown.

Screenshot 2023-11-26 at 12 59 05 AM

### DIFF
--- a/app/src/server/api/routers/datasetEntries.router.ts
+++ b/app/src/server/api/routers/datasetEntries.router.ts
@@ -1,4 +1,3 @@
-import { type ComparisonModel } from "@prisma/client";
 import { TRPCError } from "@trpc/server";
 import archiver from "archiver";
 import { sql } from "kysely";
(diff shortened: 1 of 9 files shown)


app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -780,6 +778,15 @@ export const datasetEntriesRouter = createTRPCRouter({
...
                   eb.fn.agg<number>("AVG", [`der.score`]).as(`scoreForEval_${datasetEval.id}`),
+                  sql
+                    .raw("CAST(COUNT(CASE WHEN der.score = 1 THEN 1 ELSE NULL END) AS INT)")
+                    .as(`wins_${datasetEval.id}`),
+                  sql
+                    .raw("COUNT(CASE WHEN der.score = .5 THEN 1 ELSE NULL END)")
+                    .as(`ties_${datasetEval.id}`),
+                  sql
+                    .raw("COUNT(CASE WHEN der.score = 0 THEN 1 ELSE NULL END)")
+                    .as(`losses_${datasetEval.id}`),
```
corbt: sql.raw is pretty dangerous here -- if anyone figures out how to set a custom `datasetEval.id` then we have a major SQL injection attack where they could extract pretty much anything from our database.

Can you use the sql`` template strings for this instead? That'll do the correct escaping/parameterization.

corbt (approved):
Nice, this definitely helps. I suspect we'll still need to do more to help people understand what these scores mean but this is better than before!

app/src/server/api/routers/datasetEntries.router.ts
arcticfly: You're more familiar with kysely sql syntax than I am. I think the updated code should fix the vulnerability that you identified, but I would appreciate you double-checking.

corbt: I'm still seeing a lot of sql.raw statements with templated strings -- those are the things to avoid. I think the way you're using those can hopefully all be replaced by sql`` tags.

PR #371, opened 2023-11-28
Add gpt-4-0613 and gpt-4-1106 as comparison models

arcticfly:
It's nice to see how your fine-tuned models compare to powerful GPT-4 models, and to see which GPT-4 model does the best on your data.

### Changes
* Add `GPT_4_0613` and `GPT_4_1106_PREVIEW` to ComparisonModel enum, update types
* Replace comparison model confirmation dialogue with a selection modal
* Condense model visibility logic to `useVisibleModelIds` hook
* Filter eval queries to only include data relevant to visible models
* Add jitter in retry time for generate and evaluate tasks

Comparison models modal:

Screenshot 2023-11-28 at 1 12 13 AM


app/src/server/tasks/generateTestSetEntry.task.ts
```diff
@@ -19,166 +19,194 @@ import {
...
-      const choice = completion.choices[0];
-      if (!choice) throw new Error("No completion returned");
-      const completionMessage = choice.message;
-
-      if (fineTune) {
-        const inputTokens = completion.usage?.prompt_tokens ?? 0;
-        const outputTokens = completion.usage?.completion_tokens ?? 0;
-        await prisma.usageLog.create({
+    } catch (e) {
+      console.error("error in generateTestSetEntry", e);
```
corbt, line 189: We should probably report this to Sentry?

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -108,117 +132,123 @@ export const evaluateTestSetEntries = defineTask<EvaluateTestSetEntriesJob>({
...
+          id: secondResult.id,
+        },
+        data: {
+          status: "COMPLETE",
+          explanation,
+          score: score2,
+        },
+      });
+    } catch (e) {
+      console.error("error in evaluateTestSetEntries", e);
```
corbt, line 228: Report to Sentry

corbt (approved):
Looks like a nice batch of changes!

app/src/server/tasks/generateTestSetEntry.task.ts
arcticfly: If we do that, we'll end up getting a bunch of errors due to rate limiting. We could insert a check to avoid reporting errors caused by rate limiting, but that should go in a separate PR.

PR #375, opened 2023-11-29
Add "Results" tab to evals

corbt:
A couple of simple visualizations: a table view as well as a heatmap of head to head results. Not terribly pretty atm but functional.

As I worked on this I became more convinced that we should split this into a new "Evaluations" tab and rename the current tab to "Test Set" or something like that. For one thing, it's important that we give people a way to link to and share evaluations! But didn't want to take the time to do that now.

Screenshot 2023-11-28 at 10 47 29 PM


app/src/components/datasets/DatasetContentTabs/Evaluation/ShowEvalModal/ResultsTab.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+import { useMemo } from "react";
+import ColoredPercent from "~/components/ColoredPercent";
+import { useAppStore } from "~/state/store";
+import { getOutputTitle } from "../getOutputTitle";
+
+export const ResultsTab = () => {
+  const evalId = useAppStore((state) => state.evaluationsSlice.showEvalModalId);
+  const results = api.datasetEvals.results.useQuery(
+    { datasetEvalId: evalId as string },
```
arcticfly, line 27: Should we take visible model ids into account here?

app/src/server/api/routers/datasetEvals.router.ts
```diff
@@ -469,4 +470,51 @@ export const datasetEvalsRouter = createTRPCRouter({
...
+          eb.fn.avg<number>("der.score").as("winRate"),
+          sql<number>`count(case when der.score = 1 then 1 end)::int`.as("wins"),
+          sql<number>`count(case when der.score = .5 then 1 end)::int`.as("ties"),
+          sql<number>`count(case when der.score = 0 then 1 end)::int`.as("losses"),
+        ]);
+
+      const [leaderboard, headToHead] = await Promise.all([
+        baseQuery.groupBy(["modelId1", "slug1"]).orderBy("winRate", "desc").execute(),
+
+        baseQuery
```
arcticfly, line 503: I think you could take advantage of the fact that we store the comparison output source on results and skip this join to the comparison results.

arcticfly (approved):
Awesome work! I’m ready to see this in action.

app/src/components/datasets/DatasetContentTabs/Evaluation/ShowEvalModal/ResultsTab.tsx
corbt: In this case I don't think so -- the modal interface seems divorced enough from the outputs table that it would probably be more confusing.

app/src/server/api/routers/datasetEvals.router.ts
corbt: Ah yes you're probably right. I'm too tired to handle this tonight and need it in for tomorrow so won't make this change right now, but will do so next time I touch this code (or you're free to).

PR #377, opened 2023-11-30
Remove deprecated accuracy

arcticfly:
We don't have a standard "accuracy" to show users now, so we should just point them to their evaluations.

### Changes
* Remove accuracy scores from fine-tune views
* Add more links to the dataset on which a fine-tune was trained
* Simplify a query we touched on [last night](https://github.com/OpenPipe/OpenPipe/pull/375#discussion_r1408883483)


corbt (approved):
👍

PR #379, opened 2023-11-30
Eval logic changes

corbt:
1. Randomize which output we present first to work around GPT-4's bias towards the first result.
2. Randomize the order we run evals, so partial results are more representative.
3. Replace "Alice" and "Bob" with "Response 1" and "Response 2". This does actuallys seem to make the answers slightly more correct in places where it makes a difference at all.

### DIFF
--- a/app/src/server/tasks/evaluateTestSetEntries.task.ts
+++ b/app/src/server/tasks/evaluateTestSetEntries.task.ts
@@ -13,7 +13,7 @@ import { getOpenaiCompletion } from "../utils/openai";
 import defineTask from "./defineTask";
 import { getComparisonModelName, isComparisonModel } from "~/utils/baseModels";
 import { calculateQueryDelay } from "./queryModel.task";
-import { isEqual } from "lodash-es";
+import { isEqual, shuffle } from "lodash-es";
 import { chatCompletionMessage } from "~/types/shared.types";
 
 // Accept result criteria instead of ids to recover from duplicate result creation attempts
@@ -32,8 +32,6 @@ export type EvaluateTestSetEntriesJob = {
 };
 
 const MAX_TRIES = 25;
-
-const JUDGEMENT_OPTIONS = ["ALICE_BETTER", "EQUAL", "BOB_BETTER"] as const;
 
 export const evaluateTestSetEntries = defineTask<EvaluateTestSetEntriesJob>({
   id: "evaluateTestSetEntries",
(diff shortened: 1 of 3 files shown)


app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -540,8 +538,10 @@ export const queueEvalJobsForEval = async (datasetEvalId: string) => {
   for (const datasetEvalDatasetEntry of datasetEval.datasetEvalDatasetEntries) {
     for (let i = 0; i < datasetEval.outputSources.length; i++) {
       for (let j = i + 1; j < datasetEval.outputSources.length; j++) {
-        const firstOutputSource = datasetEval.outputSources[i];
-        const secondOutputSource = datasetEval.outputSources[j];
+        const [firstOutputSource, secondOutputSource] = shuffle([
```
arcticfly: I would shuffle the order of output sources within the task itself. We're also queueing jobs in `queueHeadToHeadEvalJobsForTestingEntry`, so we would need to shuffle the order there as well if we do it before calling the task.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -582,7 +582,7 @@ export const queueEvalJobsForEval = async (datasetEvalId: string) => {
     skipDuplicates: true,
   });
 
-  for (const job of jobsToEnqueue) {
+  for (const job of shuffle(jobsToEnqueue)) {
```
arcticfly: Good idea.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -336,12 +334,12 @@ const constructJudgementInput = (
...
 ) => {
   const input: ChatCompletionCreateParams = {
-    model: "gpt-4",
+    model: "gpt-4-0613",
     messages: [
       {
         role: "system",
         content:
-          "You are an intelligent and fair judge of chatbots. Evaluate the following two chatbot responses and choose which one is better. If the outputs are of similar quality, you can mark them as EQUAL." +
+          "You are an intelligent and fair judge of LLMs. Evaluate the following two responses and choose which one is better. If the outputs are of similar quality, you can mark them as EQUAL." +
```
arcticfly: Does GPT-4-0613 have a good idea of what LLMs are?

arcticfly (approved):
Nice, I'm glad you thought of this.

app/src/server/tasks/evaluateTestSetEntries.task.ts
corbt: No, apparently doesn't. Switched this back to "chatbots".

app/src/server/tasks/evaluateTestSetEntries.task.ts
corbt: Good call, didn't realize that was being enqueued in two places.

I ended up refactoring to call the same enqueue method from both places since that'll make this kind of slip-up less likely in the future. I ended up doing the shuffling there, although in retrospect doing it within the job like you suggested probably would have worked at least as well.

corbt:
Small addition: I added a `wasFirst` field to `DatasetEvalResult`s. That way we can track exactly how big the "presented first" boost is over time.

PR #383, opened 2023-12-01
Optimize the query to create datasets from logged calls

corbt:
Unfortunately this doesn't make things much faster for our largest customers, but it keeps our RAM requirements much lower (since we aren't loading all IDs into RAM anymore) and makes dataset creation faster for projects with relatively few logged calls.


arcticfly (approved):
Thanks for taking care of this. I'm bummed that we won't be faster for our biggest customers, but fixing RAM requirements is most important.

PR #384, opened 2023-12-01
UX tweaks

corbt:
Bunch of UX tweaks, mostly to the Evaluate page.

Screenshot 2023-11-30 at 11 19 28 PM

### DIFF
--- a/app/src/components/datasets/DatasetContentTabs/Evaluation/FormattedMessage.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/Evaluation/FormattedMessage.tsx
@@ -1,66 +1,79 @@
 import { Text, VStack } from "@chakra-ui/react";
 import type { DatasetEntry } from "@prisma/client";
-import type { ChatCompletionMessageToolCall, ChatCompletionMessage } from "openai/resources/chat";
+import type { ChatCompletionMessage } from "openai/resources/chat";
 import SyntaxHighlighter from "react-syntax-highlighter";
 
 export const PotentiallyPendingFormattedMessage = ({
   output,
+  preferJson = false,
 }: {
   output: DatasetEntry["output"];
+  preferJson?: boolean;
 }) => {
   if (output) {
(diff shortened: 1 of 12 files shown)


app/src/components/datasets/DatasetContentTabs/Evaluation/FormattedMessage.tsx
```diff
@@ -1,66 +1,79 @@
...
-            style: { wordBreak: "break-all", whiteSpace: "pre-wrap" },
-          }}
-          wrapLines
-        >
-          {JSON.stringify(JSON.parse(args), null, 4)}
-        </SyntaxHighlighter>
-      </VStack>
-    );
-  }
+    formattedJson = JSON.stringify(JSON.parse(props.json), null, 2);
```
arcticfly, line 54: What happens if `json` isn't actually parseable to json? Do we render nothing?

app/src/server/api/routers/datasetEntries.router.ts
```diff
@@ -647,21 +647,21 @@ export const datasetEntriesRouter = createTRPCRouter({
...
-              `CASE
+          .orderBy(
+            () =>
+              sql`CASE
                 WHEN "averageScoreForEval"."datasetEntryId" IS NULL THEN 1
                 ELSE 0
               END`,
-            ),
           )
-          .orderBy(`averageScoreForEval.score`, sortOrder.order) as unknown as typeof baseQuery;
```
arcticfly, line 642: You're right, we didn't need that hacky typing here.

arcticfly (approved):
Makes sense 👍

app/src/components/datasets/DatasetContentTabs/Evaluation/FormattedMessage.tsx
corbt: If it isn't valid JSON it just falls back to to using the unformatted string (but still tries to highlight it as code).

PR #391, opened 2023-12-05
Replace | with Union[]

arcticfly:
Python 3.9 users should no longer have trouble with newfangled typing.

`|` was introduced as a replacement for `Union` in [PEP 604](https://peps.python.org/pep-0604/) with the release of python 3.10. Unfortunately, that means its syntax is not compatible with 3.9 versions.


corbt (approved):
Thanks! I'm assuming you tested with 3.9?

PR #392, opened 2023-12-05
Create a `stage` environment on AWS

corbt:
This PR automatically deploys a `stage` environment to AWS on each deploy to the `stage` branch. This is in preparation for moving our app to AWS from Render entirely.


app/prisma/us-west-2-bundle.pem
arcticfly, line 1: Should this be included in the repo?

infra/Pulumi.prod.yaml
```diff
@@ -0,0 +1,43 @@
+config:
+  aws-native:region: us-west-2
+  aws:region: us-west-2
+  openpipe:ANTHROPIC_API_KEY:
+    secure: AAABAG5PApLpxIrXlau1/D5VPnS7yxwv88kXog0O1T7xBxsmlcfoYWByoJ3Mw3wtM3ZXXUcrdbt/jJA5UM3Ke5Dm78lclNpoLcizDYVZ6i3+n/sRFNXXfA8musbQX6dsUAvh1Nja8mqkAp9di1IMz+eweyfHLiFPPmhrngBaD6SVc0iU9iq8dFsSPSg=
```
arcticfly, line 5: Are all these keys okay to include in the repo?

arcticfly (approved):
Thanks!

app/prisma/us-west-2-bundle.pem
corbt: Yeah this is a public key from the aws site.

Only downside to including it is that it expires in August 2024 and we'll have to remember to swap it out when that gets close or else our app will stop connecting to the DB.

infra/Pulumi.prod.yaml
corbt: Yeah they're fully encrypted, don't see a security risk here.

PR #395, opened 2023-12-06
Revert "replicate api token made optional"

corbt:
Reverts OpenPipe/OpenPipe#390

That change unfortunately broke deploys that didn't have a REPLICATE_API_TOKEN set.


arcticfly (approved):
My bad, I should have tested a deploy before approving.

PR #396, opened 2023-12-06
Run all infra on AWS

corbt:
We've successfully migrated all our infra from Render to AWS. This PR contains the infra definition files, as well as a Github Action to deploy whenever we push to `main`.

### DIFF
--- a/infra/scripts/migrate-prod-db.ts
+++ b/infra/scripts/migrate-prod-db.ts
@@ -1,0 +1,34 @@
+// Just checking this in as a historical record of the steps we took to migrate
+// the db to aws. This is not meant to be run again.
+
+import "dotenv/config";
+import { $ } from "execa";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+
+const $$ = $({ stdio: "inherit", shell: true });
+
+const oldDb = process.env.OLD_DATABASE_URL!;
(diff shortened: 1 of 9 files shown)


infra/src/app-env.ts
```diff
@@ -12,8 +12,8 @@ export const environment = new kubernetes.core.v1.Secret(
   nm("app"),
   {
     stringData: {
-      // DATABASE_URL: dbConnectionString,
-      DATABASE_URL: getSecret("DATABASE_URL"),
+      DATABASE_URL: dbConnectionString,
+      // DATABASE_URL: getSecret("DATABASE_URL"),
```
arcticfly, line 16: Do we still need this line?

infra/scripts/migrate-prod-db.ts
```diff
@@ -0,0 +1,31 @@
+import "dotenv/config";
```
arcticfly, line 4: Nit: Add comment that this file is for documentation purposes.

arcticfly (approved):
👍

infra/src/app-env.ts
corbt: One primary reason we'd be touching this code would be if we need to migrate the database again, so I'd like to leave this in as a reminder of how we did that last time.

PR #397, opened 2023-12-06
Allow sorting of dataset entries

corbt:
Make it easier to find outliers based on input/output token counts.

### DIFF
--- a/app/src/utils/hooks.ts
+++ b/app/src/utils/hooks.ts
@@ -1,13 +1,14 @@
 import { useRouter } from "next/router";
 import { type Query } from "nextjs-routes";
-import { type RefObject, useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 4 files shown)


app/src/utils/hooks.ts
```diff
@@ -201,8 +202,12 @@ export const useDatasetEntries = (refetchInterval = 0) => {
 
   const { page, pageSize } = usePageParams();
 
+  const sort =
+    useSortOrder<NonNullable<RouterInputs["datasetEntries"]["list"]["sortOrder"]>["field"]>()
```
arcticfly, line 206: I would use SortableField here.

arcticfly (approved):

app/src/utils/hooks.ts
corbt: Will refactor when we reuse this hook in the future.

PR #398, opened 2023-12-07
Add Hermes and Zephyr models

corbt:
The code organization is not great and our serialization format doesn't match that of the base models, so performance will potentially be less than ideal, particularly with small test set sizes. But good to have available for experimentation.

Screenshot 2023-12-07 at 11 40 18 AM


corbt:
Going to do this a different way.

PR #411, opened 2023-12-08
Don't try to load all the training data into memory at once

corbt:
This implements a new training data upload pattern that grabs 1000 rows from the database at a time, processes them, and then uploads them directly to Azure. That way we never have to have all of our training data in memory at once.

This is likely a pattern we'll have to implement in more places over time as datasets get larger.

### DIFF
--- a/app/src/server/tasks/fineTuning/trainFineTune.task.ts
+++ b/app/src/server/tasks/fineTuning/trainFineTune.task.ts
@@ -1,7 +1,7 @@
 import { env } from "~/env.mjs";
-import { prisma } from "~/server/db";
+import { kysely, prisma } from "~/server/db";
 import { trainerv1 } from "~/server/modal-rpc/clients";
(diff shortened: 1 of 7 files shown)


app/src/server/tasks/fineTuning/trainFineTune.task.ts
```diff
@@ -31,22 +86,39 @@ export const trainFineTune = defineTask<TrainFineTuneJob>({
   },
 });
 
+async function* iterateTrainingRows(fineTuneId: string) {
+  let offset = 0;
+  while (true) {
+    console.log("offset", offset);
```
arcticfly: Remove

arcticfly (approved):
If I'm reading this PR correctly, are we still unable to train 3.5 on massive datasets?

corbt:
> If I'm reading this PR correctly, are we still unable to train 3.5 on massive datasets?

Yep, didn't propagate the fix there yet.

PR #412, opened 2023-12-08
Properly report null json values

arcticfly:
If the user sends or OpenAI returns a null json value, we need to translate it to Prisma's own special version of null in order to save it in the database.

### DIFF
--- a/app/src/server/api/external/v1Api.router.ts
+++ b/app/src/server/api/external/v1Api.router.ts
@@ -230,10 +230,10 @@ export const v1ApiRouter = createOpenApiRouter({
             model,
             receivedAt: new Date(input.receivedAt),
             reqPayload: (input.reqPayload === null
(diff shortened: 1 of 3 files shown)


app/src/server/api/external/v1Api.router.ts
```diff
@@ -221,25 +221,36 @@ export const v1ApiRouter = createOpenApiRouter({
...
+      try {
+        await prisma.loggedCall.create({
+          data: {
+            id: newLoggedCallId,
+            projectId: ctx.key.projectId,
+            requestedAt: new Date(input.requestedAt),
+            model,
+            receivedAt: new Date(input.receivedAt),
+            reqPayload: (input.reqPayload === null
+              ? Prisma.DbNull
```
corbt: JSON null

corbt (approved):

PR #413, opened 2023-12-09
Centralize logic for copying a single dataset entry

arcticfly:
We're moving the dataset-copying logic that was duplicated in both the `relabelDatasetEntry` task and `datasetEntries.update` route into a centralized `copyEntryWithUpdates` function that's called from both contexts.

I considered a similar approach for centralizing the creation and cleanup process of many dataset entries, but decided against it. In the `importDatasetEntries` task, we update the dataset file upload status in between the steps of the cleanup process, which would be difficult to do if all three steps were called in the same function. Fortunately, since the entire cleanup process is only three function calls (`updatePruningRuleMatches`, `startDatasetTestJobs`, and `countDatasetEntryTokens`), we don't lose much by redefining it in both contexts.

### Changes
* Add `copyEntryWithUpdates` function, call in both `relabelDatasetEntry` task and `datasetEntries.update` route
* Fix typo in `relabelDatasetEntry` task name
* Relabel dataset entry `createdAt` as "Updated At", since from the user's perspective creations of new versions of dataset entries look like edits


corbt (approved):
This works.

For creation, can you at least add big comments in both places we do it reminding us not to change one without changing the other?

Longer term I expect we'll want to move to a pipeline abstraction for creation and importing of dataset entries, where instead of doing all the validation at once, then all the formatting at once, then writing all the entries at once, we go through all the steps for chunks of 1000 at a time. Otherwise we risk running out of memory as datasets get larger, as we recently saw. At that point we'll have to rewrite the progress reporting logic to track where we are in the chunk order instead of which stage of processing we've done. But not something we need to do right now.

arcticfly:
> Longer term I expect we'll want to move to a pipeline abstraction for creation and importing of dataset entries, where instead of doing all the validation at once, then all the formatting at once, then writing all the entries at once, we go through all the steps for chunks of 1000 at a time. Otherwise we risk running out of memory as datasets get larger, as we recently saw. At that point we'll have to rewrite the progress reporting logic to track where we are in the chunk order instead of which stage of processing we've done. But not something we need to do right now.

I'm going to open another PR with a first pass at a pipeline to decrease the maximum size of objects we have to process in memory at any given time.

PR #417, opened 2023-12-09
Drop LoggedCallModelResponse

arcticfly:
Should be run after #416 is deployed and run in staging and production.


app/scripts/restore-dump.ts
```diff
@@ -12,25 +12,8 @@ const argv = await yargs(hideBin(process.argv)).option("dump-file-path", {
 const dbName = "openpipe-dev";
 const templateDbName = `${dbName}_template`;
 
-const migrateTemplate = `
```
corbt, line 15: We can just delete this file I think; it was only relevant for the Render dumps that we aren't using anymore.

corbt (approved):

PR #418, opened 2023-12-11
Support python 3.8

arcticfly:
Version 4.0.4 of the python client (which hopefully people can still find) now supports python version >=3.8, <4.0.

### DIFF
--- a/client-libs/python/openpipe/test_async_client.py
+++ b/client-libs/python/openpipe/test_async_client.py
@@ -23,11 +23,9 @@ async def test_async_content():
     )
 
     last_logged = last_logged_call(client)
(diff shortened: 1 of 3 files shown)


client-libs/python/openpipe/test_async_client.py
```diff
@@ -23,11 +23,9 @@ async def test_async_content():
     )
 
     last_logged = last_logged_call(client)
+    assert last_logged.req_payload["messages"][0]["content"] == "count to 3"
```
arcticfly, line 26: Update tests to no longer expect messages to be nested under model response on the last logged call.

corbt (approved):
Nice. I'm assuming you tested in python 3.8?

PR #422, opened 2023-12-12
Hide field comparison by default

arcticfly:
We now default to only showing the head-to-head evals.

Also, I added a button at the top right of the eval tabs that takes the user to the dataset evaluations view with filters applied such that only output relevant to that eval is visible.

Finally, we now show the number of complete comparisons and poll for updated results on the Results tab until all comparisons are complete.

### DIFF
--- a/app/src/components/datasets/DatasetContentTabs/Evaluation/useVisibleEvalIds.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/Evaluation/useVisibleEvalIds.tsx
@@ -1,4 +1,4 @@
-import { useQueryParam, JsonParam, withDefault } from "use-query-params";
+import { useQueryParam, JsonParam, withDefault, encodeQueryParams } from "use-query-params";
 
 import { useDataset } from "~/utils/hooks";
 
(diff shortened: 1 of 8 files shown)


app/src/components/datasets/DatasetContentTabs/Evaluation/useVisibleEvalIds.tsx
```diff
@@ -6,62 +6,99 @@ export const EMPTY_EVALS_KEY = "_empty_";
 
 export const useVisibleEvalIds = () => {
   const dataset = useDataset().data;
-  const [visibleEvalIds, setVisibleEvalIds] = useQueryParam<string[]>(
-    "evals",
-    withDefault(JsonParam, []),
+  const [visibleEvals, setVisibleEvals] = useQueryParam<{
+    hthEvalIds: string[];
```
corbt: Can we rename to HTHEvalIds? I read through this PR like 3 times trying to figure out what "hth" was an abbreviation of until I figured out it was an acronym.

corbt (approved):

app/src/components/datasets/DatasetContentTabs/Evaluation/useVisibleEvalIds.tsx
arcticfly: I'll change it to `headToHeadEvalIds`, which is more readable and not that many more characters.

PR #426, opened 2023-12-14
Added: python QueryClient

DmitryBe:
Rationale: Access logged calls from Python to create a dataset for further processing and advanced evaluation.

Example:
```python
from openpipe.query_client import QueryClient, FilterCondition, QueryResponse

client = QueryClient.create()
res: QueryResponse = client.query_logged_calls(
    filters=[
        FilterCondition(field="chain_name", comparator="=", value="classify"),
    ]
)
len(res.calls)
```

### DIFF
--- a/app/src/server/api/external/v1Api.router.ts
+++ b/app/src/server/api/external/v1Api.router.ts
@@ -21,6 +21,9 @@ import {
 import { posthogServerClient } from "~/utils/analytics/serverAnalytics";
 import { calculateFineTuneUsageCost } from "~/utils/baseModels";
 import { createOpenApiRouter, openApiProtectedProc } from "./openApiTrpc";
+import { filtersSchema } from "~/types/shared.types";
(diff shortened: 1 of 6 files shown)


corbt:
Awesome, this is definitely functionality we want to expose through the SDK!

Hang tight though -- we're actually in the middle of swapping out our code-generated Python SDK which will affect this PR. (See https://github.com/OpenPipe/OpenPipe/pull/427).

arcticfly:
@DmitryBe The biggest changes to the API are out of the way. Would you like to fix conflicts and test this route out again?

corbt:
Also, I'm hoping that with our changes to the Python SDK generator, you won't need a separate `QueryClient` anymore? In other words, could this PR be reduced to the API changes and the downstream changes to the code-generated clients?

corbt:
Going to close as there are now significant merge conflicts here.

PR #427, opened 2023-12-15
Use fern client in python

arcticfly:
Fern generates a better SDK than our previous openapi -> python generator, which will be helpful to the users who want to start making calls on the reporting SDK itself.

We're now exporting `OpenPipeApi` and `AsyncOpenPipeApi` from the top of our package.


client-libs/python/openpipe/__init__.py
```diff
@@ -2,3 +2,5 @@
 from openai import *
 from .openai_sync_wrapper import OpenAIWrapper as OpenAI
 from .openai_async_wrapper import AsyncOpenAIWrapper as AsyncOpenAI
+
+from .api_client_fern.client import OpenPipeApi, AsyncOpenPipeApi
```
arcticfly, line 6: In another PR we should rename `api_client_fern` to `api_client`. I don't want to do that in this change because it would be confusing to track what files were added/deleted.

client-libs/python/openpipe/shared.py
```diff
@@ -5,66 +5,62 @@
...
-        configured_client.token = os.environ["OPENPIPE_API_KEY"]
+        client._client_wrapper._token = os.environ["OPENPIPE_API_KEY"]
 
     if os.environ.get("OPENPIPE_BASE_URL"):
-        configured_client._base_url = os.environ["OPENPIPE_BASE_URL"]
+        client._client_wrapper._base_url = os.environ["OPENPIPE_BASE_URL"]
 
-    if openpipe_options and "verify_ssl" in openpipe_options:
-        configured_client._verify_ssl = bool(openpipe_options["verify_ssl"])
+    # if openpipe_options and "verify_ssl" in openpipe_options:
```
arcticfly: I'm not sure what the equivalent of `verify_ssl` is with Fern. We should figure that out.

corbt: We might not need this, I think it was just because when running against localhost the client would complain that SSL wasn't configured. If Fern isn't complaining about that might be fine.

corbt (approved):
Ok yeah the generated SDK here is a lot better.

Follow up for this we should rename to api_client like you said and make the changes to the `report` endpoint (make the timestamps optional at a minimum) and let the customer who needed direct API access know it's cleaner now.

PR #429, opened 2023-12-15
Make requestedAt and receivedAt optional

arcticfly:
Users who want to upload request logs manually don't always have accurate `requestedAt` and `receivedAt` timestamps.

### DIFF
--- a/app/src/server/api/external/v1Api.router.ts
+++ b/app/src/server/api/external/v1Api.router.ts
@@ -195,6 +195,9 @@ export const v1ApiRouter = createOpenApiRouter({
     )
     .output(z.object({ status: z.union([z.literal("ok"), z.literal("error")]) }))
     .mutation(async ({ input, ctx }) => {
+      // Zod default messes up the generated OpenAPI spec, so we do it manually
+      if (!input.requestedAt) input.requestedAt = Date.now();
+
       const reqPayload = await reqValidator.spa(input.reqPayload);
       const respPayload = await respValidator.spa(input.respPayload);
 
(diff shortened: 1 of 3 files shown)


corbt (commented):
I think we should make both requestedAt and receivedAt optional.

Also, I think this could be slightly more clean with zod's [`default`](https://github.com/colinhacks/zod?tab=readme-ov-file#default) operator. Eg. 

```ts
requestedAt: z.number().optional().default(() => new Date()).describe("Unix timestamp in milliseconds")
```

(haven't tried it but think it should work)

arcticfly:
> I think we should make both requestedAt and receivedAt optional.
> 
> Also, I think this could be slightly more clean with zod's [`default`](https://github.com/colinhacks/zod?tab=readme-ov-file#default) operator. Eg.
> 
> ```ts
> requestedAt: z.number().optional().default(() => new Date()).describe("Unix timestamp in milliseconds")
> ```
> 
> (haven't tried it but think it should work)

Let's make `requestedAt` optional as we discussed. However, I won't use the zod default here, since openapi actually executes whatever callback we provide when generating openapi.json, resulting in a router spec that looks like this:

```
"properties": {
  "requestedAt": {
    "type": "number",
    "default": 1702684884414,
    "description": "Unix timestamp in milliseconds"
  },
```

corbt (approved):
👍

Shouldn't this change require regenerating the python and typescript client SDKs as well?

arcticfly:
> 👍
> 
> Shouldn't this change require regenerating the python and typescript client SDKs as well?

The typescript SDK already was generated (see DefaultService.ts) but I don't think the python SDK actually has any changes.

PR #430, opened 2023-12-16
Add Mistral Optimized

corbt:
This PR adds our new Mistral Optimized model, which will be described and benchmarked in a forthcoming blog post!


app/src/server/fineTuningProviders/openpipe/types.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "zod";
```
corbt, line 1: I will remove all the models except our existing ones and `OpenPipe/mistral-ft-optimized-1215` in a future follow-up. They won't actually appear in the UI and I need them in the codebase for now because otherwise the analysis script I'm using to generate data for my blog post breaks.

app/prisma/schema.prisma
```diff
@@ -362,27 +362,32 @@ enum FineTuneStatus {
...
+    provider             FineTuneProvider
+    baseModel            String
+    status               FineTuneStatus   @default(PENDING)
     errorMessage         String?
     trainingBlobName     String?
     trainingStartedAt    DateTime?
     trainingFinishedAt   DateTime?
     deploymentStartedAt  DateTime?
     deploymentFinishedAt DateTime?
-    inferenceUrls        String[]       @default([])
```
arcticfly, line 385: Don't we still need inferenceUrls?

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
```diff
@@ -61,6 +66,10 @@ const FineTuneButton = () => {
 
 export default FineTuneButton;
 
+const visibleModels = getEntries(supportedModels)
+  .filter(([_, model]) => model.trainable)
+  .map(([id, _]) => id) as [ProviderWithModel, ...[ProviderWithModel]];
```
arcticfly, line 71: You could use AtLeastOne here.

app/src/server/api/routers/fineTunes.router.ts
```diff
@@ -47,7 +47,7 @@ export const fineTunesRouter = createTRPCRouter({
       });
 
       return {
-        fineTunes,
+        fineTunes: fineTunes.map(typedFineTune),
```
arcticfly, line 50: Why do we need to type our fineTunes? Shouldn't they already be typed when pulled from the database?

app/src/server/fineTuningProviders/supportedModels.ts
```diff
@@ -0,0 +1,141 @@
...
+    cost: sevenBCosts,
+    trainable: false,
+  },
+  "openpipe:OpenPipe/mistral-7b-1214-gamma": {
+    name: "Mistral 7B Optimized (1214g)",
+    cost: sevenBCosts,
+    trainable: false,
+  },
+
+  "openai:gpt-3.5-turbo-0613": {
```
arcticfly, line 101: Nit: group openai models in list

app/src/server/fineTuningProviders/supportedModels.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const sevenBCosts = {
+  training: 0.000004,
+  input: 0.0000012,
+  output: 0.0000016,
+};
+
+// Do not remove models from this list if any users still have them in production.
+// Instead, mark them as `trainable: false` so no one else can create new models.
+export const supportedModels: Record<ProviderWithModel, FrontendModelInfo> = {
```
arcticfly, line 21: Does the order in which models are declared in this object affect the order in which they are rendered in the UI?

app/src/server/fineTuningProviders/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const baseModel = z.discriminatedUnion("provider", [
+  z.object({ provider: z.literal("openai"), baseModel: openaiModels }).passthrough(),
+  z.object({ provider: z.literal("openpipe"), baseModel: openpipeModels }).passthrough(),
+]);
+
+export type BaseModel = z.infer<typeof baseModel>;
+
+// Gives us a type of eg "openai:gpt-3.5-turbo-1106" | "openpipe:OpenPipe/meta-llama/Llama-2-7b-hf"
+export type ProviderWithModel = BaseModel extends infer T
```
arcticfly, line 13: I didn't know generating dynamic string types from object types was possible. Did GPT-4 help with this?

trainer/deploy-modal-apps.sh
```diff
@@ -7,7 +7,7 @@ set -e
 cd "$(dirname "$0")"
 
 # Default environment
-ENV="prod"
+ENV="dev"
```
arcticfly, line 10: How do the prod and dev environments differ on Modal?

app/prisma/schema.prisma
arcticfly: Ah, I forgot that you're using `huggingFaceModelId` to determine which model to load.

arcticfly (approved):
I'm looking forward to using the new model!

app/src/server/api/routers/fineTunes.router.ts
corbt: `typedFineTune` ensures that the `model` is one of the supported models for the given provider, so we don't eg. try to get the model name + costs for some random model string that doesn't exist and break the UI. Instead we crash very loudly here. 😁

app/src/server/fineTuningProviders/supportedModels.ts
corbt: I'm grouping models by trainable/not-trainable. So since we no longer support training of turbo-0613 I moved it to the bottom.

app/src/server/fineTuningProviders/supportedModels.ts
corbt: Yep, exactly. They'll be ordered in the UI in the same order they appear in this file.

app/src/server/fineTuningProviders/types.ts
corbt: I knew it was possible but yes, GPT-4 taught me how!

trainer/deploy-modal-apps.sh
corbt: Theoretically they don't, but when we're testing in dev don't want to be pushing live changes to prod. Once I deploy this PR I'll need to push the updated training code to the Modal prod environment.

PR #431, opened 2023-12-16
Accept azure messages in chat completion route

arcticfly:
The cleanest way to handle Azure's omission of the `content` field in function call messages is to clean assistant messages by setting content to null when we process the message. We're already doing this when converting request logs to dataset entries, but we also need to do this when accepting assistant messages as input to our `/chat-completions` route.

### DIFF
--- a/app/src/server/api/external/v1Api.router.ts
+++ b/app/src/server/api/external/v1Api.router.ts
@@ -12,10 +12,9 @@ import { prisma } from "~/server/db";
 import {
   chatCompletionInputReqPayload,
   chatCompletionOutput,
-  chatMessageAzure,
+  chatMessage,
   functionCallInput,
   functionsInput,
-  normalizeAzureMessage,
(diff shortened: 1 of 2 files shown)


corbt (commented):
Actually I think just adding `.default(null)` to the end of [this line](https://github.com/OpenPipe/OpenPipe/blob/add-models-take-2/app/src/types/shared.types.ts#L96) might do the trick?

arcticfly:
> Actually I think just adding `.default(null)` to the end of [this line](https://github.com/OpenPipe/OpenPipe/blob/add-models-take-2/app/src/types/shared.types.ts#L96) might do the trick?

Yes, I suppose there is _that_ option.

corbt (approved):

corbt:
Btw I think that fix might make the ugly check I created to set this value to `null` as part of importing dataset entries redundant as well, so maybe we can remove that?

PR #433, opened 2023-12-16
Fix test set entry generation for GPT-3.5 fine-tunes

corbt:
My recent change accidentally broke generating test-set entries for GPT-3.5 fine tunes. This fixes that, and also tightens up the typing for fine tunes in general. Tightening the typing is actually what led me to discover the bug!


corbt:
Going to merge directly once I've done a smoke test locally since currently GPT-3.5 FTs are broken in prod.

arcticfly:
Thanks for catching this.

PR #435, opened 2023-12-18
Sort fine-tunes just by date

corbt:
A user got confused because their newest fine tune was showing up low on the list because it was for an old dataset. Let's just sort the fine-tunes page by fine-tune date; if they want to see them grouped by dataset they can go to the datasets page.


arcticfly (approved):

arcticfly:
That's what I get for trying to be too clever.

PR #436, opened 2023-12-18
Allow user to apply only certain pruning rules to a fine tune

arcticfly:
It's useful to compare performance between models both with and without pruning rules. We'll now allow users to determine which pruning rules should be applied to the model they want to train.

Training modal:
Screenshot 2023-12-18 at 4 04 37 PM

### DIFF
--- a/app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
@@ -102,9 +102,14 @@ const FineTuneModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
     if (disclosure.isOpen) {
       setSelectedBaseModel(visibleModels[0]);
       setModelSlug(humanId({ separator: "-", capitalize: false }));
-      setAppliedPruningRuleIds(pruningRules?.map((rule) => rule.id) ?? []);
     }
-  }, [disclosure.isOpen, pruningRules]);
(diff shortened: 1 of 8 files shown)


corbt:
Initial UX feedback from the screenshot:
 - I like putting the pruning rules inside a frame, but making the frame alert-colored feels weird. Maybe just a thin gray border and white background?
 - How hard would it be to have a "Show" / "Hide" button next to each rule, eg. "[ ] Rule #1 [Show]". That would quickly jog my memory on which rule I'm enabling/disabling without having to open a new tab or navigate back and forth while keeping the rule numbers in my short-term memory.

corbt (approved):
See my previous comment for UX feedback. Code looks good.

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
```diff
@@ -225,7 +230,7 @@ const FineTuneModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
                 >
                   <HStack>
                     <Text fontWeight="bold">Applied Pruning Rules</Text>{" "}
-                    <InfoCircle tooltipText="Use pruning rules to save tokens when sending inputs to your fine-tuned model." />
+                    <InfoCircle tooltipText="Use pruning rules to reduce the number of input token your fine-tuned model has to process." />
```
corbt: "input tokens"

PR #437, opened 2023-12-19
Add Vanta role

corbt:
New IAM role that lets Vanta perform a security audit of our infra.

### DIFF
--- a/infra/index.ts
+++ b/infra/index.ts
@@ -6,6 +6,7 @@ import "./src/dns";
 export * from "./src/metrics";
 export { dbConnectionString } from "./src/database";
 export { deployAccessKeyId, deployAccessKeySecret } from "./src/deployUser";
+export * from "./src/vanta";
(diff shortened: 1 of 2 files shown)


infra/index.ts
```diff
@@ -6,6 +6,7 @@ import "./src/dns";
 export * from "./src/metrics";
 export { dbConnectionString } from "./src/database";
 export { deployAccessKeyId, deployAccessKeySecret } from "./src/deployUser";
+export * from "./src/vanta";
```
arcticfly, line 9: For whose benefit is this being exported?

arcticfly (approved):

infra/index.ts
corbt: That was just so I could see the role ARN to copy into the Vanta UI. I could have gotten that from the AWS web dashboard as well though I guess.

PR #438, opened 2023-12-19
Show pruning rules in fine-tune general tab

arcticfly:
After pruning rules have been applied to a fine-tuned model, users need a way to view them.

Fine tune general tab:
Screenshot 2023-12-18 at 5 44 13 PM


corbt (approved):
Nice work

PR #442, opened 2023-12-19
Record which model was used as judge

arcticfly:
Since we're now using both gpt-4-0613 and gpt-4-1106-turbo as judges, it's useful to keep track of which model was used in each evaluation.


app/scripts/copy-db.sh
```diff
@@ -58,7 +58,16 @@ create_db() {
 
 run_update_sql_on_temp_db() {
     report_progress "Running SQL update on temporary database..."
-    psql -d "$TEMP_DB_NAME" -c 'update "DatasetEntry" set "loggedCallId" = null;'
+    # Account for some logged calls that were not properly copied over
```
corbt: revert

corbt (approved):

PR #444, opened 2023-12-20
Add API docs to Mintlify

arcticfly:
A couple customers have asked to see our API docs, so let's add them in Mintlify.

Additionally, let's host our openapi.json in our public folder to make it easy to download.

Swagger:
Screenshot 2023-12-19 at 11 32 36 PM


corbt:
The swagger UI is ugly and I'm a little worried about their code quality (this should not add a thousand lines to our lockfile!) Can we just use mintlify instead?

arcticfly:
> The swagger UI is ugly and I'm a little worried about their code quality (this should not add a thousand lines to our lockfile!) Can we just use mintlify instead?

Yep, Mintlify's version does look better.

corbt (approved):

corbt (commented):
The OpenAPI spec is already hosted at https://app.openpipe.ai/api/v1/openapi.json which feels like a more natural place for it anyway (since it's specific to the v1 public API). Do we need it in public/ as well?

arcticfly:
> The OpenAPI spec is already hosted at https://app.openpipe.ai/api/v1/openapi.json which feels like a more natural place for it anyway (since it's specific to the v1 public API). Do we need it in public/ as well?

No we don't. Let's keep it there.

PR #448, opened 2023-12-21
Allow streaming from the /chat/completions route

arcticfly:
We now allow streaming from our /chat/completions route for base and fine-tuned OpenAI models. We don't yet support streaming for fine-tuned Mistral models, but that should be coming shortly.

Additionally, this PR adds support for logging calls as part of the request to /chat/completions, instead of requiring an additional call to the /report route.


app/src/server/api/external/v1Api.router.ts
```diff
@@ -105,8 +93,8 @@ export const v1ApiRouter = createOpenApiRouter({
           .optional(),
       }),
     )
-    .output(chatCompletionOutput)
-    .mutation(async ({ input, ctx }): Promise<ChatCompletion> => {
+    .output(z.union([chatCompletionOutput.nullable(), z.any()]))
```
arcticfly, line 97: I'm not sure of the right way to type a ReadableStream in zod. z.any works in the short term, but we should come up with a better solution.

corbt: I don't think it's possible. Pretty sure Zod is just focused on typing things that are JSON-serializable, which a readable stream definitely isn't.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
+        }
+      }
 
-        return completion;
-      } catch (error: unknown) {
-        console.error(error);
-        throw new TRPCError({
-          message: `Failed to get completion: ${(error as Error).message}`,
-          code: "BAD_REQUEST",
+      const logRequest = ctx.headers["op-log-request"] === "true";
```
corbt: I think we should default to reporting unless you explicitly opt out. That makes it much simpler for us to tell someone "if you want to start using openpipe with [your random openai wrapper library], just use our baseUrl" instead of them also having to figure out how to set a custom header with their wrapper lib, which may not even be possible.

To avoid double-reporting from our SDKs in the short-term we can only default to `true` if you're calling a non-openpipe model. But we should also update our SDKs to set this to `true` and stop calling the reporting endpoint separately.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
+          tags = JSON.parse(ctx.headers["op-tags"] as string);
+        } catch (error: unknown) {
+          // ignore
+        }
+      }
+
+      if (completion instanceof Stream) {
+        // split stream to avoid locking the read mutex
+        const [recordStream, outputStream] = completion.tee();
+        void captureRecordUsage({
```
corbt: Shouldn't this go inside `recordUsage`?

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
-      } catch (error: unknown) {
-        console.error(error);
-        throw new TRPCError({
-          message: `Failed to get completion: ${(error as Error).message}`,
-          code: "BAD_REQUEST",
+      const logRequest = ctx.headers["op-log-request"] === "true";
+      let tags: Record<string, string> = {};
+      if (ctx.headers["op-tags"]) {
+        try {
+          tags = JSON.parse(ctx.headers["op-tags"] as string);
```
corbt, line 162: Let's actually just throw an error if you pass in an `op-tags` instance but it isn't formatted correctly. That's better than silently swallowing it and the user not knowing if they set the wrong header name, or misformatted it, or something else.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
+          inputPayload,
+          completion: recordStream,
+          logRequest,
+          fineTune,
+          tags,
         });
+        return outputStream.toReadableStream();
       }
+
+      void recordUsage({
```
corbt: Let's add a `.catch(e => captureException(e))` or similar to all of these so see in Sentry if these are breaking. We should strive to never silently swallow errors.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
+        try {
+          tags = JSON.parse(ctx.headers["op-tags"] as string);
+        } catch (error: unknown) {
+          // ignore
+        }
+      }
+
+      if (completion instanceof Stream) {
+        // split stream to avoid locking the read mutex
+        const [recordStream, outputStream] = completion.tee();
```
corbt, line 175: Nice, I should use `tee` more.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -34,27 +36,42 @@ export async function getCompletion2(
...
-        completion.choices[0].message,
-      ) as ChatCompletionMessage;
-    }
     return completion;
-  } else {
-    return getModalCompletion(fineTune, prunedInput);
   }
+
+  // TODO: create pipeline without this conversion once OpenAI supports tool_calls for their fine-tuned models
+  input = convertToolCallInputToFunctionInput(input);
```
corbt: I'm a little confused about what's going on here. Why are we converting tool calls only for non-streamed responses? Shouldn't tool calls be treated the same way independent of whether you're streaming or not?

app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -21,7 +23,7 @@ import { type TypedFineTune } from "~/types/dbColumns.types";
 export async function getCompletion2(
```
corbt, line 25: So I realize it looks ugly but maybe we want to add multiple function declarations here? That way typescript knows that if we pass in stream: true, we get a stream, and if not not. Eg:

```ts
export async function getCompletion2(
  fineTune: TypedFineTune,
  input: ChatCompletionCreateParamsNonStreaming,
): Promise<ChatCompletion>;
export async function getCompletion2(
  fineTune: TypedFineTune,
  input: ChatCompletionCreateParamsStreaming,
): Promise<Stream<ChatCompletionChunk>>;
export async function getCompletion2(
  fineTune: TypedFineTune,
  input: ChatCompletionCreateParams,
): Promise<ChatCompletion | Stream<ChatCompletionChunk>>;
```

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
-        where: { slug: modelSlug },
-      });
-
-      if (!fineTune) {
-        throw new TRPCError({ message: "The model does not exist", code: "NOT_FOUND" });
-      }
+      const fineTune =
+        (await prisma.fineTune.findUnique({
+          where: { slug: modelSlug },
+        })) ?? undefined;
```
corbt: What's the `?? undefined` for?

app/src/server/api/external/v1Api.router.ts
```diff
@@ -115,62 +103,89 @@ export const v1ApiRouter = createOpenApiRouter({
...
 
-      const typedFT = typedFineTune(fineTune);
-
-      if (typedFT.projectId !== key.projectId) {
+      if (inputPayload.model.startsWith("openpipe:") && !fineTune) {
```
corbt: The flow of this function feels somewhat hard to follow with me. I think it would be cleaner to just have an `if (inputPayload.model.startswith("openpipe:"))` at the top and split at that point, instead of having some checks here and then a separate `if (fineTune)` down below.

app/src/server/tasks/generateTestSetEntry.task.ts
```diff
@@ -125,21 +125,22 @@ export const generateTestSetEntry = defineTask<GenerateTestSetEntryJob>({
...
         response_format: datasetEntry.response_format ?? undefined,
+        stream: false,
       };
 
       try {
         if (isComparisonModel(modelId)) {
           completion = await getOpenaiCompletion(rawDatasetEntry.dataset.projectId, input);
         } else if (fineTune) {
-          completion = await getCompletion2(fineTune, input);
+          completion = (await getCompletion2(fineTune, input)) as ChatCompletion;
```
corbt: If we make the change to the typings of `getCompletion2` we can drop this. If we don't though should make this an actual check on the return type, not a type assertion. Too easy to make make mistakes with type assertions!

app/src/server/api/external/v1Api.router.ts
corbt: Also we need to use zod to validate this is actually a record of strings.

app/src/utils/recordRequest.ts
```diff
@@ -0,0 +1,239 @@
...
+import { posthogServerClient } from "./analytics/serverAnalytics";
+import {
+  countLlamaInputTokens,
+  countLlamaOutputTokens,
+  countOpenAIChatTokens,
+} from "./countTokens";
+import { default as openAIModelProvider } from "~/modelProviders/openai-ChatCompletion";
+import { type BaseModel } from "~/server/fineTuningProviders/types";
+
+export const captureRecordUsage = async ({
```
corbt: This is basically a wrapper for `recordUsage` that works with streaming, right? Can't we just let `recordUsage` take a completion of type stream or chatcompletion and handle this internally? A separate wrapper I feel like just adds indirection for not much gain in this case.

app/src/server/api/external/v1Api.router.ts
corbt: Also let's put this inside an `else`, so it's clear that this is stuff that happens if we're *not* streaming. I missed that on my first read-through and got confused.

app/src/server/api/external/v1Api.router.ts
arcticfly: Sure. I think that should go in a separate PR as a larger refactor of our SDKs to allow streaming (currently our SDKs will break if you try to use the streaming features this PR enables for fine-tuned models).

app/src/server/api/external/v1Api.router.ts
arcticfly: I don't see a problem with combining recordUsage and captureRecordUsage. Let's do it.

app/src/server/api/external/v1Api.router.ts
arcticfly: Setting it to undefined makes our calls to `recordUsage` a little cleaner. Hopefully the refactor clarified its purpose.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
arcticfly: I added a comment earlier in the file to clarify. Basically converting a streamed function call into a tool call output on the fly would be complicated.

app/src/modelProviders/fine-tuned/getCompletion-2.ts
arcticfly: It's not ugly. It's beautiful.

app/src/server/api/external/v1Api.router.ts
arcticfly: I refactored it

corbt (approved):
great work here.

PR #452, opened 2023-12-23
Use filters to select training data

arcticfly:
Let's let users train on subsets of their dataset using filters. Let's also give some indication of how sufficient the size of their training dataset is.

Training Size Gauge:
Screenshot 2023-12-22 at 9 25 15 PM

### DIFF
--- a/app/src/pages/datasets/[id]/[tab].tsx
+++ b/app/src/pages/datasets/[id]/[tab].tsx
@@ -1,11 +1,10 @@
-import { Breadcrumb, BreadcrumbItem, Center, Flex, Icon, Input, VStack } from "@chakra-ui/react";
+import { Breadcrumb, BreadcrumbItem, Center, Flex, Icon, VStack, Text } from "@chakra-ui/react";
 import Link from "next/link";
-import { useState, useEffect } from "react";
+import { useEffect } from "react";
 import { AiOutlineDatabase } from "react-icons/ai";
 
 import AppShell from "~/components/nav/AppShell";
-import { api } from "~/utils/api";
-import { useDataset, useHandledAsyncCallback } from "~/utils/hooks";
+import { useDataset } from "~/utils/hooks";
 import PageHeaderContainer from "~/components/nav/PageHeaderContainer";
 import ProjectBreadcrumbContents from "~/components/nav/ProjectBreadcrumbContents";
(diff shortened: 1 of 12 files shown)


app/src/pages/datasets/[id]/[tab].tsx
```diff
@@ -67,20 +46,7 @@ export default function Dataset() {
...
-                borderWidth={1}
-                borderColor="transparent"
-                fontSize={16}
-                px={0}
-                minW={{ base: 100, lg: 300 }}
-                flex={1}
-                _hover={{ borderColor: "gray.300" }}
-                _focus={{ borderColor: "blue.500", outline: "none" }}
-              />
+              <Text>{dataset.data?.name}</Text>
```
Comment (corbt) [inline app/src/pages/datasets/[id]/[tab].tsx:49]: Yeah I like this change, we should be consistent in the way we rename things.

app/src/server/api/routers/fineTunes.router.ts
```diff
@@ -180,6 +182,23 @@ export const fineTunesRouter = createTRPCRouter({
         input.baseModel.baseModel,
       );
 
+      await kysely
```
corbt, line 185: How long does this operation take with a 100k row dataset? I understand why you moved it here and it definitely simplifies the code, but I'm worried we might end up providing a pretty bad UX if this takes 20+ seconds with large datasets.

Eventually I think we probably will have to move it back to the background, since I expect we'll end up supporting 1M+ row datasets. But don't have to optimize for that use case yet.

arcticfly: I’m not sure, let me test how long that takes. I do think that in most cases creating the training entries as part of the initial fine-tune creation handler results in a better UX. This way when you navigate to the fine-tunes page, training entries is not initially set to 0, which is confusing and feels buggy. I would much rather wait 20 seconds and know that it worked.

arcticfly: Creating a fine-tune with about 80k training rows (for a dataset with 100k rows in total) took about 8 seconds locally. It didn't feel like a long wait, and I was happy to immediately see the training set size when I navigated to the fine-tunes page.

corbt (approved):
Ok I was playing around with this locally and changed some of the wording. If that looks good to you I'm good to merge.

PR #453, opened 2023-12-25
Add copiable code block to ft page

arcticfly:
It's nice to be able to try out your fine-tuned model without digging through the docs on running inference. Let's include a copiable code block that provides runnable code in `cURL`, `Python`, and `Javascript` form.

Fine Tune General tab:
Screenshot 2023-12-27 at 2 25 33 PM

### DIFF
--- a/app/src/components/fineTunes/FineTuneContentTabs/General/General.tsx
+++ b/app/src/components/fineTunes/FineTuneContentTabs/General/General.tsx
@@ -92,9 +92,11 @@ const General = () => {
         <FineTunePruningRules />
         <FineTuneDangerZone />
       </VStack>
-      <Box position="sticky" top={8} pl={{ base: 0, md: 8 }}>
-        <InferenceCodeTabs />
-      </Box>
+      {fineTune.status === "DEPLOYED" && (
+        <Box position="sticky" top={8} pl={{ base: 0, md: 8 }}>
+          <InferenceCodeTabs />
+        </Box>
+      )}
     </Flex>
   );
 };
--- a/app/src/components/fineTunes/FineTuneContentTabs/General/General.tsx
+++ b/app/src/components/fineTunes/FineTuneContentTabs/General/General.tsx
@@ -1,4 +1,4 @@
-import { VStack, HStack, Text, Button, Heading } from "@chakra-ui/react";
+import { VStack, HStack, Text, Button, Heading, Box, Flex } from "@chakra-ui/react";
 import { useFineTune, useHandledAsyncCallback } from "~/utils/hooks";
 import ContentCard from "~/components/ContentCard";
 import FineTuneDangerZone from "./FineTuneDangerZone";
@@ -9,6 +9,7 @@ import ViewEvaluationButton from "~/components/datasets/DatasetContentTabs/Evalu
 import ViewDatasetButton from "~/components/datasets/ViewDatasetButton";
 import { modelInfo } from "~/server/fineTuningProviders/supportedModels";
 import FineTunePruningRules from "./FineTunePruningRules";
+import InferenceCodeTabs from "./InferenceCodeTabs/InferenceCodeTabs";
 
 const General = () => {
   const fineTune = useFineTune().data;
@@ -25,8 +26,15 @@ const General = () => {
   if (!fineTune) return null;
 
   return (
-    <VStack w="full" h="full">
-      <VStack w="full" alignItems="flex-start" spacing={4} pb={12}>
+    <Flex
+      h="fit-content"
+      position="relative"
+      w="full"
+      alignItems="flex-start"
+      pb={12}
+      flexDir={{ base: "column", md: "row" }}
+    >
+      <VStack w="full" alignItems="flex-start" spacing={4}>
         <ContentCard>
           <VStack w="full" alignItems="flex-start" spacing={4} bgColor="white">
             <Heading size="md" pb={4}>
@@ -84,7 +92,10 @@ const General = () => {
         <FineTunePruningRules />
         <FineTuneDangerZone />
       </VStack>
-    </VStack>
+      <Box position="sticky" top={8} pl={{ base: 0, md: 8 }}>
+        <InferenceCodeTabs />
+      </Box>
+    </Flex>
   );
 };
 
--- a/app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/CopiableCodeTabs.tsx
+++ b/app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/CopiableCodeTabs.tsx
@@ -1,0 +1,127 @@
+import { useState } from "react";
+import {
+  Tabs,
+  TabList,
+  Tab,
+  TabPanels,
+  TabPanel,
+  HStack,
+  IconButton,
+  useToast,
+  Text,
+  Tooltip,
+} from "@chakra-ui/react";
+import SyntaxHighlighter from "react-syntax-highlighter";
+import { tomorrowNightBright } from "react-syntax-highlighter/dist/cjs/styles/hljs";
+import { CopyIcon } from "lucide-react";
+
+export type CodeTab = {
+  title: string;
+  code: string;
+  language: string;
+};
+
+const CopiableCodeTabs = ({ tabs }: { tabs: CodeTab[] }) => {
+  const [tabIndex, setTabIndex] = useState(0);
+
+  const toast = useToast();
+
+  const copyToClipboard = async (text?: string) => {
+    try {
+      await navigator.clipboard.writeText(text as string);
+      toast({
+        title: "Copied to clipboard",
+        status: "success",
+        duration: 2000,
+      });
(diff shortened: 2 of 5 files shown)


app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/CopiableCodeTabs.tsx
```diff
@@ -0,0 +1,127 @@
...
+  code: string;
+  language: string;
+};
+
+const CopiableCodeTabs = ({ tabs }: { tabs: CodeTab[] }) => {
+  const [tabIndex, setTabIndex] = useState(0);
+
+  const toast = useToast();
+
+  const copyToClipboard = async (text?: string) => {
```
corbt: We use this in enough places that we should probably extract it into a component that takes the content to be copied as a prop and renders the copy icon.

app/src/components/fineTunes/FineTuneContentTabs/General/General.tsx
```diff
@@ -84,7 +92,10 @@ const General = () => {
         <FineTunePruningRules />
         <FineTuneDangerZone />
       </VStack>
-    </VStack>
+      <Box position="sticky" top={8} pl={{ base: 0, md: 8 }}>
```
corbt: I think we should only render this conditionally if the model is actually deployed? It would be confusing to see this code block, try to run my model, and get an error that it isn't ready yet.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
```diff
@@ -0,0 +1,131 @@
...
+"model": "openpipe:{{FINE_TUNE_MODEL_SLUG}}",
+"messages": {{TEMPLATED_MESSAGES}},
+"tool_choice": {{TEMPLATED_TOOL_CHOICE}},
+"tools": {{TEMPLATED_TOOLS}}
+}'`,
+  },
+  {
+    title: "Python",
+    language: "python",
+    code: `from openpipe import OpenAI
```
corbt: Let's add a comment to the top for both this and the JS one where we explain how to install the library. Eg. `# pip install openpipe`

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const fineTune = useFineTune().data;
+  const datasetEntries = useTrainingEntries().data;
+
+  const templatedTabs = useMemo(() => {
+    const entry = datasetEntries?.entries[0];
+
+    return templateTabs(project?.openpipeApiKey || "", fineTune?.slug || "", entry?.datasetEntry);
+  }, [project?.openpipeApiKey, fineTune?.slug, datasetEntries?.entries]);
+
+  return <CopiableCodeTabs tabs={templatedTabs} />;
```
corbt, line 132: I'm a little confused by the code organization here. Why is this a separate nested component? They seem very strongly linked.

corbt (commented):
This is a nice UX improvement. I like the idea of making the code example a sidebar, but the theming here feels really "tacked on" next to the rest of the page and our app in general. Some specific changes that should help it feel more cohesive:
 - put the whole thing inside a card with 0 body margin, so we get a consistent border and border radius that matches the rest of our panels.
 - Change the theme to light to match the rest of the app (can just use the same code highlighting we have for function outputs on the evaluate tab probably.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const formattedToolChoice = JSON.stringify(entry?.tool_choice ?? "auto", null, 4);
+
+  const formattedTools = JSON.stringify(entry?.tools ?? []);
+
+  return baseTabs.map((tab) => {
+    const templatedCode = tab.code;
+
+    const tabFormattedMessages =
+      tab.language === "bash"
+        ? formattedMessages.replace(/'/g, `'"'"'`)
```
corbt, line 100: If we start running into issues where some edge cases aren't escaped properly might be a good idea to switch to a dedicated library for this like https://www.npmjs.com/package/shell-escape.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
arcticfly: Because there is already a ton of logic in the top-level component, and I like to limit component size and complexity.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/CopiableCodeTabs.tsx
arcticfly: We only use this copy+toast functionality in two places, and the styles are different between those two. I'm happy to create a shared component that accepts different styles, but I'm not sure that it's a win from a code-complexity perspective.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
arcticfly: I don't think we will, but if we do I'd prefer to use another package. This one would require us to write our own types, and isn't very ergonomic.

arcticfly:
> This is a nice UX improvement. I like the idea of making the code example a sidebar, but the theming here feels really "tacked on" next to the rest of the page and our app in general. Some specific changes that should help it feel more cohesive:
> 
> * put the whole thing inside a card with 0 body margin, so we get a consistent border and border radius that matches the rest of our panels.
> * Change the theme to light to match the rest of the app (can just use the same code highlighting we have for function outputs on the evaluate tab probably.

This is definitely a question of taste, but I actually like the contrast. It's similar to how Mintlify, OpenAI, and other established companies have displayed code on their sites. I know you like light styles in your code editor, but most people use a dark mode, as do I.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/CopiableCodeTabs.tsx
arcticfly: I experimented with putting both copy icons in the same component, but styling with a tooltip became too complex, so I decided to just put `copyToClipboard` in its own hook.

corbt: I see it in `FormattedJson`, `CopiableCode` and here? If the formatting is different enough then that's fair, can make it a hook instead of a component.

corbt: ah just saw your comment, sg!

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
corbt: Encapsulating logic makes a ton of sense when it's reused, and can make sense even without reuse if it presents a clean and intuitive interface so the caller doesn't have to understand the inner workings of the callee. However, in this case realistically if you're working in these tabs you're going to have to understand both components anyway, so splitting them up doesn't really gain us much. In fact it's the opposite -- I found myself having to bounce back and forth between these files several times to figure out which pieces are defined where.

corbt:
> This is definitely a question of taste, but I actually like the contrast. It's similar to how Mintlify, OpenAI, and other established companies have displayed code on their sites. I know you like light styles in your code editor, but most people use a dark mode, as do I.

Ok, we can leave it dark. Let's still match the border radius to cards, and use a color that's closer to what's already in our palette like blue.300 or something for the tab highlights so it feels more cohesive with the rest of our site.

app/src/components/fineTunes/FineTuneContentTabs/General/InferenceCodeTabs/InferenceCodeTabs.tsx
arcticfly: It makes sense that you had to flit between both of these files to understand their combined functionality when first reading through the code. I think we should continue to separate the logic for determining what to display from the logic for displaying it to cut down on the amount of logic in each file - otherwise we'll end up with an instance of that brand of enormous monolithic component that makes development so unpleasant.

corbt (approved):

PR #455, opened 2023-12-29
update docs with logging recommendations

bjsi:
Just adding some logging recommendations based on a discussion with @corbt


corbt (approved):
Thanks @bjsi!

PR #456, opened 2023-12-30
Add slugs to projects

arcticfly:
Let's add a project slug to our project-specific urls. This change will be useful when linking to public projects from our docs.


app/@types/nextjs-routes.d.ts
```diff
@@ -22,21 +22,21 @@ declare module "nextjs-routes" {
...
-    | DynamicRoute<"/fine-tunes/[id]/[tab]", { "id": string; "tab": string }>
-    | DynamicRoute<"/fine-tunes/[id]", { "id": string }>
-    | StaticRoute<"/fine-tunes">
     | StaticRoute<"/">
     | DynamicRoute<"/invitations/[invitationToken]", { "invitationToken": string }>
-    | StaticRoute<"/project/settings">
-    | StaticRoute<"/request-logs">
-    | StaticRoute<"/sentry-example-page">
-    | StaticRoute<"/usage">;
+    | DynamicRoute<"/p/[slug]/datasets/[id]/[tab]", { "slug": string; "id": string; "tab": string }>
```
corbt: I think this route segment should probably be called [project_slug] instead of just [slug]? Could get confusing if we have a nested resource that also has a slug.

app/next.config.mjs
```diff
@@ -29,6 +29,44 @@ let config = {
...
+    },
+    {
+      source: "/usage",
+      destination: "/",
+      permanent: true,
+    },
+    {
+      source: "/p/:path",
+      destination: "/p/:path/request-logs",
+      permanent: true,
```
corbt: Let's not make this permanent; I could imagine us wanting to change the default route in the future and don't want users' browsers to cache the /request-logs path forever just in case.

app/prisma/migrations/20231229214948_add_project_slug/migration.sql
```diff
@@ -0,0 +1,57 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[slug]` on the table `Project` will be added. If there are existing duplicate values, this will fail.
+
+*/
+
+-- Replace the current nanoid function with a reduced alphabet and length of 8
+CREATE OR REPLACE FUNCTION nanoid(
+    size int DEFAULT 8,
```
corbt: chatgpt did some math and with these values our collision probability is actually higher than I thought (16% in the first 1M projects). Let's extend the length to 10 and include uppercase letters to mitigate that.

corbt (commented):
The ProjectLink abstraction seems like a clean one!

app/@types/nextjs-routes.d.ts
arcticfly: I renamed it to be `projectSlug`, because in general I prefer camelCase when working in javascript.

corbt (approved):

app/prisma/schema.prisma
```diff
@@ -167,6 +167,7 @@ model PruningRuleMatch {
 
 model Project {
     id   String @id @default(uuid()) @db.Uuid
+    slug String @unique() @default(dbgenerated("nanoid()")) @db.VarChar(10)
```
corbt, line 170: I should have commented on this earlier, sorry for not catching. In Postgres there's no efficiency advantage to a `VarChar` over the standard `text` type, the only thing it does is add a constraint. I generally prefer to just stick to `text` for everything given that as it's more flexible and better supported. Not worth changing at this point though imo.

PR #458, opened 2023-12-31
Add public projects with viewers

arcticfly:
Let's add the concept of a public project that can be viewed by anyone but only modified by admins and members. Access control checks will prevent viewers from modifying the project in any way, but a read-only API token will allow them to experiment with the project's models.

### Changes
* Add read-only API key
* Prevent viewers with read-only API tokens from recording request logs
* Add AccessControl checks to UI
* Allow project admins to invite viewers
* Prevent viewers from seeing other viewers
* Allow projects to be public and automatically add viewers
* Allow user to leave project
* Order projects by the date the user joined

Sample restricted action:
Screenshot 2023-12-31 at 1 58 04 AM

Member API key view:
Screenshot 2024-01-01 at 8 49 48 AM

Viewer API key view:
Screenshot 2024-01-01 at 8 50 34 AM


app/src/components/datasets/DatasetContentTabs/Evaluation/AddEvalModal.tsx
```diff
@@ -231,17 +233,23 @@ const AddEvalModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
...
+                colorScheme="blue"
+                onClick={onCreationConfirm}
+                minW={24}
+                isLoading={creationInProgress}
+                isDisabled={
+                  !name ||
+                  !instructions ||
+                  !numDatasetEntries ||
+                  includedModelIds.length < 2 ||
+                  insufficientPermission
```
arcticfly: Unfortunately, when buttons possess their own `isDisabled` prop, AccessControl doesn't override it, so we need to manually disable them. When buttons don't define `isDisabled` for themselves, AccessControl automatically disables them when appropriate.

corbt: Just thought of another possible abstraction here that will actually lead to a better UX and also solve this problem more cleanly. What if we generalized the access control wrapper component into something like this:

```ts
<ConditionallyEnable
  accessRequired="requireCanViewProject"
  checks={[
    [selectedRows > 10, "You must select at least 10 rows before starting a fine-tune"],
    [modelName === 'gpt-3.5' || hasBetaAccess, "You need beta access before you can fine-tune this model"]
    // ...etc
  ]}
>
// ...
```

Basically, the idea is to move all the `isDisabled` checks into the higher-level access control wrapper. That way we can also easily add a tooltip for each reason we might disable the button, which will provide a better UX.

corbt (approved):
Nice work!

PR #460, opened 2024-01-01
Access control tweaks

corbt:
Ok so first of all sorry for just kinda doing a drive-by on your PR. I had checked it out to experiment with ways to get more type safety into the hook/component for when we add more args besides `projectId` and it sort of spiraled from there. I didn't actually make any improvements on that front, but I did make some other changes that I think are useful:

1. Pull the `accessDeniedText` from the `TRPCError` message. This means we'll have better messaging when we deny API actions and can reuse that messaging in the user-facing buttons as well.
2. Rename `accessLevel` to `accessCheck`. I think this is a more intuitive name for the concept since access checks aren't really well-ordered as levels.

### DIFF
--- a/app/src/utils/accessControl.ts
+++ b/app/src/utils/accessControl.ts
@@ -11,6 +11,16 @@ const isAdmin = async (userId: string) => {
   return !!user;
 };
 
+const requireUserId = (ctx: TRPCContext) => {
+  if (!ctx.session?.user.id) {
+    throw new TRPCError({
+      code: "UNAUTHORIZED",
+      message: "This function is only available to signed-in users.",
+    });
+  }
+  return ctx.session.user.id;
+};
+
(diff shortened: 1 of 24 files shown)


app/src/utils/accessControl.ts
```diff
@@ -47,18 +54,18 @@ export const requireCanViewProject = async (projectId: string, ctx: TRPCContext)
...
+      });
     }
   }
 };
 
 export const requireCanModifyProject = async (projectId: string, ctx: TRPCContext) => {
   ctx.markAccessControlRun();
 
-  const userId = ctx.session?.user.id;
-  if (!userId || !projectId) {
```
arcticfly, line 59: This check is actually important. If userId or projectId are not set, then the prisma.projectUser.findFirst call will error. Please add it back in to all the access check functions.

arcticfly (changes requested):

PR #461, opened 2024-01-01
Set and track training configs app-side

corbt:
This change makes our TS app responsible for generating the full Axolotl config for a training run, instead of the Python trainer. There are a few reasons for this:
 - It lets us more easily experiment with different training configs in development.
 - It will let us allow end-users to more easily override specific hparams (eg. number of epochs, learning rate, etc.) in the future.
 - It makes it easier for us to charge for fine-tuning, since we're now generating the number of epochs app-side and can accurately compute the number of training tokens from the app.

I'm working on this right now because we need to test different hparams as part of figuring out the right balance between quality and throughput for LoRA hot-swapping.


app/src/server/api/internal/v1Api.router.ts
```diff
@@ -25,25 +25,17 @@ export const v1ApiRouter = createOpenApiRouter({
       z.object({
         trainingDataUrl: z.string(),
         huggingFaceModelId: z.string(),
-        baseModel: z.string(),
-        projectName: z.string(),
-        modelSlug: z.string(),
+        trainingConfig: z.record(z.unknown()),
```
arcticfly: Why isn't this known?

app/src/server/fineTuningProviders/openpipe/trainingConfig.ts
```diff
@@ -0,0 +1,146 @@
...
+  lr_scheduler: "cosine",
+  learning_rate: 0.0002,
+  train_on_inputs: false,
+  group_by_length: false,
+  gradient_checkpointing: true,
+  warmup_steps: 10,
+  weight_decay: 0.0,
+  optimizer: "adamw_bnb_8bit",
+  output_dir: "__placeholder__",
+
```
arcticfly: Any reason for this newline?

arcticfly (approved):

app/src/server/fineTuningProviders/openpipe/trainingConfig.ts
corbt: Removed

app/src/server/api/internal/v1Api.router.ts
corbt: I hadn't generated the full axolotl config as a zod schema, but I decided it was worth the hassle, so this isn't `unknown` anymore.

PR #465, opened 2024-01-03
Experiments with LoRA rank

corbt:
Benchmarking performance and response quality with different LoRA ranks on disparate tasks.


arcticfly (approved):
Looks good. Results are promising.

PR #466, opened 2024-01-03
Properly stringify and report unusual tag types

arcticfly:
Let's allow users to send tags of type boolean, number, and null.


app/src/modelProviders/fine-tuned/getCompletion-2.ts
```diff
@@ -129,7 +129,7 @@ async function getModalCompletion(
   return {
     id: resp.id,
     object: "chat.completion",
-    created: Date.now(),
+    created: Math.round(Date.now() / 1000),
```
arcticfly, line 132: As part of testing this change, I noticed that OpenAI returns the current date in seconds, not milliseconds.

corbt (approved):
Looks right.

PR #467, opened 2024-01-03
Update Python SDK, reformat errors

arcticfly:
This PR does 3 big things:

1. Update the Python SDK to use the latest version of OpenAI, and to call our /chat/completions route with logging enabled.
2. Include a new `error` object in the errors that we throw from our /chat/completions route so that the OpenAI TypeScript SDK shows good error messages when an exception is thrown.
3. Ensure the request is always logged, even if an error is thrown.


app/package.json
```diff
@@ -107,7 +107,7 @@
     "recharts": "^2.7.2",
     "stream-buffers": "^3.0.2",
     "superjson": "1.12.2",
-    "trpc-openapi": "github:OpenPipe/trpc-openapi#build",
```
arcticfly, line 111: Related changes to openapi-trpc: https://github.com/OpenPipe/trpc-openapi/compare/build...OpenPipe:trpc-openapi:extended-trpc-build?expand=1

corbt (approved):

PR #472, opened 2024-01-04
Persist selected project

arcticfly:
Persist the last project that the user selected for the next time they navigate to the base url.


corbt (approved):
Nice, this will make things feel more intuitive.

PR #473, opened 2024-01-04
Use S-LoRA for inference

corbt:
This PR switches us to the S-LoRA branch of vLLM and starts serving future fine-tunes via S-LoRA, instead of by serving the fully-merged model.

It does however maintain backwards compatibility. After training a model, we still persist the merged model to disk, and we could switch to serving it via the old process just by setting the fine-tune's pipeline version to 2 instead of 3.

There is an (intentional) regression here: for new pipeline v3 models trained in prod, we'll no longer be able to automatically serve them in development. I'll need to write a script that we can call to explicitly copy production models into development when necessary for testing.


app/package.json
```diff
@@ -20,13 +20,14 @@
...
     "codegen:clients": "tsx scripts/codegen-clients.ts",
+    "codegen:modal": "tsx src/server/modal-rpc/codegen-modal.ts",
     "codegen:db": "prisma generate && kysely-codegen --dialect postgres --out-file src/types/kysely-codegen.types.ts && prettier . --write",
     "seed": "NODE_OPTIONS='--no-node-snapshot' tsx prisma/seed.ts",
     "check": "concurrently 'pnpm lint' 'pnpm tsc' 'pnpm prettier . --check'",
     "test": "pnpm vitest",
     "repl": "tsx scripts/repl.ts",
     "repl:watch": "tsx --watch scripts/repl.ts",
-    "dev:modal": "cd ../trainer && MODAL_ENVIRONMENT='dev' concurrently 'poetry run modal serve src/inference_server/main.py' 'poetry run modal serve src/trainer/main.py' --kill-others"
+    "dev:modal": "cd ../trainer && MODAL_ENVIRONMENT='dev' concurrently 'poetry run modal serve src/lora_inference/main.py' 'poetry run modal serve src/trainer/main.py' --kill-others"
```
arcticfly, line 30: Does `lora_inference/main.py` also set up the serving of existing merged models?

app/src/modelProviders/fine-tuned/getCompletion.ts
```diff
@@ -51,7 +45,14 @@ export async function getCompletion2(
   if (fineTune.provider === "openai") {
     return getOpenAIFineTuneCompletion(fineTune, prunedInput);
   } else {
-    return getModalCompletion(fineTune, prunedInput);
+    switch (fineTune.pipelineVersion) {
+      case 1:
```
arcticfly, line 49: Could be worth commenting what each of these pipeline versions is. I'm starting to lose track.

trainer/src/lora_inference/main.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return output
+
+
+web_app = fastapi.FastAPI(title=APP_NAME)
+
+
+@stub.function(
+    allow_concurrent_inputs=500,
+    timeout=1 * 60 * 60,
```
arcticfly, line 152: Is the timeout now an hour?

trainer/src/shared.py
```diff
@@ -47,3 +49,8 @@ def upload_directory_to_s3(local_directory, destination, bucket):
 
         for future in as_completed(futures):
             future.result()
+
+
+def require_auth(auth_credentials=Depends(HTTPBearer())):
```
arcticfly, line 54: 👏

arcticfly (approved):
Very excited to try this out!x

app/src/modelProviders/fine-tuned/getCompletion.ts
corbt: Added comments now

app/package.json
corbt: No, those will continue to be served from the endpoint defined in inference_server/main.py

PR #476, opened 2024-01-04
Capture training usage logs

arcticfly:
We've been seeing our training usage ramp up more quickly than we anticipated, so let's keep track of how many dataset entries are being trained in the form of usage logs. Let's also backfill previous training in backdated usage logs.


app/src/server/scripts/backfillTrainingUsageLogs.ts
```diff
@@ -0,0 +1,100 @@
...
+      .select(() => [
+        sql<number>`sum(ftte.prunedInputTokens)`.as("totalInputTokens"),
+        sql<number>`sum(de.outputTokens)`.as("totalOutputTokens"),
+      ])
+      .executeTakeFirst();
+
+    const totalInputTokens = trainingStats?.totalInputTokens ?? 0;
+    const totalOutputTokens = trainingStats?.totalOutputTokens ?? 0;
+
+    await prisma.usageLog.create({
```
arcticfly, line 143: @corbt Somewhere in here we need to multiply the number of tokens by the number of epochs. How do we know how many epochs we trained on?

corbt: https://github.com/OpenPipe/OpenPipe/blob/main/app/src/server/fineTuningProviders/openpipe/trainingConfig.ts#L87

app/src/server/scripts/backfillTrainingUsageLogs.ts
```diff
@@ -0,0 +1,100 @@
...
+
+console.log("Found fineTunes to backfill:", fineTunesToBackfill.length);
+
+for (let i = 0; i < fineTunesToBackfill.length; i++) {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const typedFT = typedFineTune(fineTunesToBackfill[i]!);
+  console.log(`Backfilling fineTune (${i + 1}/${fineTunesToBackfill.length}):${typedFT.id}`);
+
+  const stringsToPrune = await getStringsToPrune(typedFT.id);
+  const isOpenAi = typedFT.baseModel.startsWith("gpt-3.5");
```
corbt: I think you can just look at `typedFT.provider`

app/src/server/tasks/fineTuning/checkFineTuneStatus.task.ts
```diff
@@ -33,23 +35,45 @@ export const checkFineTuneStatus = defineTask({
...
+              ])
+              .executeTakeFirst();
+
+            const totalInputTokens = trainingStats?.totalInputTokens ?? 0;
+            const totalOutputTokens = trainingStats?.totalOutputTokens ?? 0;
+
+            await prisma.usageLog.create({
+              data: {
+                fineTuneId: typedFT.id,
+                inputTokens: totalInputTokens,
```
corbt, line 67: Remember to multiply by epoch here as well.

corbt (approved):
Nice. I do think it's worth adding a column to the fine-tune to track the number of epochs it was trained for; I expect sooner or later someone will want to be able to set that explicitly and I can also imagine wanting to show that in the UI.

PR #483, opened 2024-01-06
Add route for adding, updating, and deleting logged call tags

arcticfly:
To improve dataset quality, users should be able to add and edit logged call tags. Users can now update and delete tags on their logged calls by passing in an array of filters with values that should be equal to or contained in the provided fields.


app/src/server/api/external/v1Api.router.ts
```diff
@@ -75,12 +46,8 @@ export const v1ApiRouter = createOpenApiRouter({
       },
     })
     .input(
-      // TODO: replace this whole mess with just `chatCompletionInput` once
-      // no one is using the `reqPayload` field anymore.
+      // TODO: replace this whole mess with just `chatCompletionInputReqPayload`
       z.object({
```
corbt, line 92: Have we checked no one is using the old sdk versions? I'm not sure what Axis is on for example.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -334,6 +301,149 @@ export const v1ApiRouter = createOpenApiRouter({
...
+        protect: true,
+      },
+    })
+    .input(
+      z.object({
+        filters: z
+          .union([
+            z.object({
+              field: z.union([z.enum(["model", "completionId"]), z.string()]),
+              equals: z.union([z.string(), z.number(), z.boolean()]),
```
corbt: If we're only matching the "model" and "completionId" field shouldn't the allowed value only be of type `string`?

app/src/server/api/external/v1Api.router.ts
```diff
@@ -334,6 +301,149 @@ export const v1ApiRouter = createOpenApiRouter({
 
       return { status: "ok" };
     }),
+  updateLogTags: openApiProtectedProc
+    .meta({
+      openapi: {
+        method: "POST",
+        path: "/update-log-tags",
```
corbt: Perhaps "/logs/update-tags" might be the better route here?

app/src/server/api/external/v1Api.router.ts
arcticfly: I searched the db for recent request logs that contained reqPayload and came up empty. We deprecated this field 3 months ago, and I think most of our users have updated since then.

app/src/server/api/external/v1Api.router.ts
arcticfly: We also match tags.

app/src/server/api/external/v1Api.router.ts
arcticfly: No strong opinion on my part, I'll update it.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -334,6 +301,149 @@ export const v1ApiRouter = createOpenApiRouter({
...
+    })
+    .input(
+      z.object({
+        filters: z
+          .union([
+            z.object({
+              field: z.union([z.enum(["model", "completionId"]), z.string()]),
+              equals: z.union([z.string(), z.number(), z.boolean()]),
+            }),
+            z.object({
```
corbt: I think we had discussed only supporting `equals` for the moment and waiting for someone to ask for this? With our current db setup it would be trivial to DoS us if you start relying on `contains` and calling it in a loop.

arcticfly: Okay, I can remove contains.

corbt:
Would you mind squashing this PR into a single commit? I often look at the git blame for a given line to remind myself why things are the way they are as I'm going through the code, and since this PR currently deletes and restores the deprecated route code in two separate commits it would break that workflow.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -107,6 +110,14 @@ export const v1ApiRouter = createOpenApiRouter({
           ? chatCompletionInputReqPayload.parse(input.reqPayload)
           : chatCompletionInputReqPayload.parse(input);
 
+      if ("reqPayload" in input) {
```
corbt, line 122: Let's do this for `checkCache` as well.

app/src/server/api/external/v1Api.router.ts
```diff
@@ -334,6 +347,135 @@ export const v1ApiRouter = createOpenApiRouter({
...
+            projectId: ctx.key.projectId,
+            loggedCallId,
+          })),
+        );
+      }
+
+      if (dataToInsert.length) {
+        await kysely
+          .insertInto("LoggedCallTag")
+          .columns(["name", "value", "projectId", "loggedCallId"])
```
corbt, line 482: This is actually just my ignorance about kysely since I assume this works, but why do you set `id: uuidv4()` above if it isn't one of the columns you're inserting into here?

app/src/server/api/external/v1Api.router.ts
```diff
@@ -334,6 +347,135 @@ export const v1ApiRouter = createOpenApiRouter({
 
       return { status: "ok" };
     }),
+  updateLogTags: openApiProtectedProc
```
corbt, line 361: It's your call since you've done the local testing, but personally this is the kind of method where I'd probably feel much more comfortable writing some tests to make sure all the edge cases around updating/deleting/adding are properly covered, especially since we might easily update the behavior of `parseTags` or `constructLoggedCallFiltersQuery` in the future for something frontend-related and forget it applies here as well.

app/src/server/utils/parseTags.ts
```diff
@@ -1,6 +1,8 @@
 import { z } from "zod";
 
-export const parseTags = (tags: unknown) => {
+export function parseTags(tags: unknown): Record<string, string>;
+export function parseTags(tags: unknown, keepNulls: true): Record<string, string | null>;
+export function parseTags(tags: unknown, keepNulls?: true) {
   if (typeof tags !== "object") {
```
corbt, line 6: Seems like some of these checks could be removed/simplified by starting with `const parsedTags = z.record(z.union([z.string(), z.number(), z.boolean(), z.null()]))` at the top here?

app/src/server/api/external/v1Api.router.ts
arcticfly: That is a valid question. Empirically, if I don't include an id for each tag, this operation fails. However, I'm not sure why this operation works while "id" is excluded from the enumerated columns into which we are inserting.

app/src/server/api/external/v1Api.router.ts
arcticfly: Do you think the checks in the TypeScript sdk omit obvious edge cases?

app/src/server/utils/parseTags.ts
arcticfly: I made the decision not to use zod for some of these checks because I'm not a big fan of the error messages it produces. I regret that we automated a large part of our dataset entry upload parsing pipeline with zod, since I think it's a lot harder to figure out what's actually going on for a first-time user.

app/src/server/api/external/v1Api.router.ts
corbt: Fair, they're pretty good. The main ones that come to mind are (1) one that doesn't update *any rows* and (2) one that updates 2+ rows (but explicitly doesn't update a third row that matches some but not all of the filters). I'd also inspect the non-updated rows to make sure they aren't actually updated -- I realize that isn't really a thing we can do easily from the SDK side though.

app/src/server/utils/parseTags.ts
corbt: Fair, for certain kinds of errors we can definitely give more guidance manually than you'd get out of Zod. Although Zod actually lets you set custom error messages when a property fails validation, which might get us a lot of the way there while keeping it simplicity. https://github.com/colinhacks/zod/blob/master/ERROR_HANDLING.md#schema-bound-error-map

arcticfly:
> Would you mind squashing this PR into a single commit? I often look at the git blame for a given line to remind myself why things are the way they are as I'm going through the code, and since this PR currently deletes and restores the deprecated route code in two separate commits it would break that workflow.

Screenshot 2024-01-07 at 5 54 07 PM

I'm not sure what you mean, I think I always squash and merge.

app/src/server/api/external/v1Api.router.ts
arcticfly: I added those tests to the sdk.

corbt (approved):
Awesome work

PR #488, opened 2024-01-08
Add indexes to speed up queries

arcticfly:
Listing fine tune training entries and assembling a list of logged call tags for a project should be much faster with these changes.


corbt (approved):
Looks good!

PR #491, opened 2024-01-09
Bump the current_pipeline version

corbt:
I think we've seen at this point that pipeline 3 is performing well -- let's make that the default for new models going forward.


arcticfly (approved):
👏

PR #492, opened 2024-01-09
Fall back to azure gpt4

arcticfly:
Temporary fix for handling remarkably high loads as we scale up our infrastructure.


app/prisma/schema.prisma
```diff
@@ -390,6 +390,8 @@ model FineTune {
     deploymentFinishedAt   DateTime?
     numEpochs              Float?
     numTrainingAutoretries Int              @default(0)
+    useCache               Boolean          @default(false)
```
corbt: Not using this right?

corbt (approved):
It's a band-aid but better than not having one.

app/prisma/schema.prisma
arcticfly: Correct, it's now removed.

PR #497, opened 2024-01-11
Implement basic response caching

arcticfly:
Let's save our users some money by allowing them to cache their old responses.


corbt:
A few questions:
1. How do you feel about an `op-options` header that would be a JSON object with both `cache` and `tags` keys? One advantage would be that our system would be more predictable for a new user: basically anything you can add in the `openpipe:` key in the SDK you can also add to the `op-options` header.
2. How should this interact with usage logs? I'm tentatively thinking we should still generate a log for each cache hit, with the correct number of input/output tokens, but set the cost to 0, and set it to a different type (eg `type=CACHE_HIT`) and hide it from all aggregations on the usage page for now. Would be good to have around though (1) for our own analytics, (2) to expose to the user at some future point, and (3) because we might eventually want to charge some small amount for this. I guess one downside here is that `fineTuneId` isn't always set, if the cache hit is for an OpenAI model.
3. We probably do need to ship API endpoints to remove a single cache hit by completion id, as well as clear the cache for a project entirely.

arcticfly:
> A few questions:
> 
> 1. How do you feel about an `op-options` header that would be a JSON object with both `cache` and `tags` keys? One advantage would be that our system would be more predictable for a new user: basically anything you can add in the `openpipe:` key in the SDK you can also add to the `op-options` header.

I actually like splitting our options into separate headers, which I believe is the standard. Writing multiple options in a json object is more cumbersome than specifying them individually.

> 2. How should this interact with usage logs? I'm tentatively thinking we should still generate a log for each cache hit, with the correct number of input/output tokens, but set the cost to 0, and set it to a different type (eg `type=CACHE_HIT`) and hide it from all aggregations on the usage page for now. Would be good to have around though (1) for our own analytics, (2) to expose to the user at some future point, and (3) because we might eventually want to charge some small amount for this. I guess one downside here is that `fineTuneId` isn't always set, if the cache hit is for an OpenAI model.

I agree, we should generate usage logs for cache hits but set the cost to 0. I created a new `CACHE_HIT` UsageType as you suggested.

> 3. We probably do need to ship API endpoints to remove a single cache hit by completion id, as well as clear the cache for a project entirely.

Agreed, but let's ship this first and let our customers start saving right away. We can add the API endpoint in a separate PR.

corbt (approved):
Good enough for now!

PR #499, opened 2024-01-12
Upsert new cache entries

arcticfly:
Upserting new cache entries will avoid rare race conditions in which the same cache entry is added from two different requests simultaneously.

### DIFF
--- a/app/src/server/api/external/v1Api.router.ts
+++ b/app/src/server/api/external/v1Api.router.ts
@@ -271,8 +271,10 @@ export const v1ApiRouter = createOpenApiRouter({
             tags,
           }).catch((e) => captureException(e));
           if (useCache && !cachedCompletion) {
-            await prisma.cachedResponse.create({
-              data: {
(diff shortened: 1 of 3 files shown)


corbt (approved):

corbt:
lol looks like we're still getting errors we try to update the cache with multiple simultaneous requests... because prisma's upsert appears to be misleadingly named and doesn't always use the db upsert, so it still has an internal race condition. I think the right fix here is to switch to kysely for this query where we can do a real db upsert that's guaranteed not to conflict. https://openpipe.sentry.io/issues/4859274485/?project=4505642011394048&referrer=github-pr-bot https://www.prisma.io/docs/orm/reference/prisma-client-reference#database-upsert-query-criteria

corbt:
@arcticfly would you mind fixing this by moving to kysely for the upsert? I'm still seeing these errors in prod and it's likely showing up as 500s to users.

PR #508, opened 2024-01-15
Correctly redirect to default project settings page from docs

arcticfly:
When we updated how our project urls worked, we neglected to account for incoming links from docs.


Kovbo (approved):
It looks good

PR #509, opened 2024-01-15
Point to updated generic project settings link

arcticfly:
Let's update our docs to point to the new generic project settings link.


Kovbo (approved):
Nice, it should work fine.

PR #512, opened 2024-01-16
Invitations rate limit implemented

Kovbo:
Now, we have an invitation rate limit of 10 invites per hour.

Email-ProjectId constraint is removed on the DB level in order to track sent invitations.

How it works:
-If you click "Cancel", it sets "isCanceled" to true in the DB.
-The invited users cannot accept an invitation if it is canceled.
-If you invite a user again, it says an error. To change a role, a user needs to cancel an invitation and send a new one.
-A user cannot get/see canceled invitations.

Additionally, added loggedCallId index and onDelete: SetNull

### DIFF
--- a/app/src/server/api/routers/users.router.ts
+++ b/app/src/server/api/routers/users.router.ts
@@ -210,7 +210,7 @@ export const usersRouter = createTRPCRouter({
         },
       });
 
-      if (!invitation) {
+      if (!invitation || invitation.isCanceled) {
         throw new TRPCError({ code: "NOT_FOUND" });
       }
 
--- a/app/src/server/api/routers/users.router.ts
+++ b/app/src/server/api/routers/users.router.ts
@@ -1,7 +1,6 @@
 import { v4 as uuidv4 } from "uuid";
 import { z } from "zod";
 import { TRPCError } from "@trpc/server";
-
 import { createTRPCRouter, protectedProcedure } from "~/server/api/trpc";
 import { prisma } from "~/server/db";
 import { error, success } from "~/utils/errorHandling/standardResponses";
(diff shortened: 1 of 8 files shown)


app/src/server/api/routers/users.router.ts
```diff
@@ -120,7 +138,7 @@ export const usersRouter = createTRPCRouter({
         },
       });
 
-      if (!invitation) {
+      if (!invitation || invitation.isCanceled) {
```
arcticfly, line 141: I would add isCanceled = false to the database queries. Then you can get rid of this check after the fact.

app/src/server/api/routers/users.router.ts
```diff
@@ -42,21 +56,25 @@ export const usersRouter = createTRPCRouter({
         });
 
         if (existingMembership) {
-          return error(`A user with ${input.email} is already invited to this project`);
+          return error(`A user ${input.email} is already accepted an invitation`);
         }
       }
 
-      const invitation = await prisma.userInvitation.upsert({
+      //Checking if a user is already invited to the project
```
arcticfly, line 63: Nit: add space between `//` and the rest of the text.

app/src/server/api/routers/users.router.ts
```diff
@@ -31,6 +30,21 @@ export const usersRouter = createTRPCRouter({
...
+            gte: new Date(Date.now() - 1 * 60 * 60 * 1000), // 1 hour
+          },
+        },
+      });
+
+      if (recentInvitationsCount > 10) {
+        return error("Invitation Limit Exceeded: Please wait before sending more invites.");
+      }
+
+      //Checking is a user is already a member of the project
```
arcticfly, line 47: Nit: "Checking is" -> "Checking if"

corbt:
Great work!

PR #515, opened 2024-01-16
Add loggedCallId index on DatasetEntry

arcticfly:
This should speed up deleting request logs.


corbt (approved):
Honestly thought this would take minutes or hours but cool, glad it doesn't!

PR #516, opened 2024-01-16
Retry relabel task up to 30 times

arcticfly:
Previously, we didn't retry relabeling tasks after they ran into an error. Now we retry up to 25 times with exponential backoff.


corbt (approved):
Good catch.

PR #517, opened 2024-01-17
Order by role before createdAt

arcticfly:
Project membership level is more important than when someone joined a project, so we should give it greater cardinality.


Kovbo (approved):
Nice solution!

PR #519, opened 2024-01-17
Add docs on evaluations

arcticfly:
We should let users know how evaluations work and what they're used for.


corbt (approved):
Thanks for this!

PR #520, opened 2024-01-18
Stats UI updated

Kovbo:
UI change: Stats are displayed by months (by default) and separated into two tabs: cost and activity.
API change: Stats API accepts startDate and endDate.

### DIFF
--- a/app/src/components/dashboard/CustomTooltip.tsx
+++ b/app/src/components/dashboard/CustomTooltip.tsx
@@ -1,0 +1,39 @@
+import { Card, Text, VStack } from "@chakra-ui/react";
+import { toUTCDateShortFormat } from "~/utils/dayjs";
+
+interface PayloadItem {
+  name: string;
+  color: string;
+  value: string;
+}
+
+function CustomTooltip({
+  active,
+  payload,
+  label,
+  beforeValue,
+}: {
+  active?: boolean;
+  payload?: PayloadItem[];
+  label?: string;
+  beforeValue?: string;
+}) {
+  if (active && payload && payload.length && label) {
+    return (
+      <Card>
+        <VStack alignItems="flex-start" p={4}>
+          <Text as={"b"}>{toUTCDateShortFormat(label)} UTC</Text>
+          {payload.map((p) => (
+            <Text key={p.name} style={{ color: p.color }}>{`${p.name}: ${beforeValue || ""}${
+              p.value
+            }`}</Text>
+          ))}
+        </VStack>
+      </Card>
+    );
+  }
+
+  return <></>;
+}
+
+export default CustomTooltip;
--- a/app/src/components/dashboard/CustomTooltip.tsx
+++ b/app/src/components/dashboard/CustomTooltip.tsx
@@ -11,12 +11,12 @@ function CustomTooltip({
   active,
   payload,
   label,
-  beforeValue,
+  valuePrefix,
 }: {
   active?: boolean;
   payload?: PayloadItem[];
   label?: string;
-  beforeValue?: string;
+  valuePrefix?: string;
 }) {
   if (active && payload && payload.length && label) {
     return (
@@ -24,7 +24,7 @@ function CustomTooltip({
         <VStack alignItems="flex-start" p={4}>
           <Text as={"b"}>{formatToUTCDayMonth(label)} UTC</Text>
           {payload.map((p) => (
-            <Text key={p.name} style={{ color: p.color }}>{`${p.name}: ${beforeValue || ""}${
+            <Text key={p.name} style={{ color: p.color }}>{`${p.name}: ${valuePrefix || ""}${
               p.value
             }`}</Text>
           ))}
--- a/app/src/components/dashboard/CustomTooltip.tsx
+++ b/app/src/components/dashboard/CustomTooltip.tsx
@@ -1,5 +1,5 @@
 import { Card, Text, VStack } from "@chakra-ui/react";
-import { toUTCDateShortFormat } from "~/utils/dayjs";
+import { formatToUTCDayMonth } from "~/utils/dayjs";
 
 interface PayloadItem {
   name: string;
@@ -22,7 +22,7 @@ function CustomTooltip({
     return (
       <Card>
         <VStack alignItems="flex-start" p={4}>
-          <Text as={"b"}>{toUTCDateShortFormat(label)} UTC</Text>
+          <Text as={"b"}>{formatToUTCDayMonth(label)} UTC</Text>
           {payload.map((p) => (
             <Text key={p.name} style={{ color: p.color }}>{`${p.name}: ${beforeValue || ""}${
               p.value
--- a/app/src/pages/_app.tsx
+++ b/app/src/pages/_app.tsx
@@ -43,7 +43,8 @@ const MyApp: AppType<{ session: Session | null }> = ({
           <SyncAppStore />
           <Favicon />
           <ChakraThemeProvider>
-            <QueryParamProvider adapter={NextAdapterPages}>
+            {/* enableBatching allows to update 2 states at once. It is crucial for the Usage page. */}
+            <QueryParamProvider adapter={NextAdapterPages} options={{ enableBatching: true }}>
               <Component {...pageProps} />
             </QueryParamProvider>
           </ChakraThemeProvider>
--- a/app/src/pages/_app.tsx
+++ b/app/src/pages/_app.tsx
@@ -44,7 +44,7 @@ const MyApp: AppType<{ session: Session | null }> = ({
           <Favicon />
           <ChakraThemeProvider>
             {/* enableBatching allows to update 2 states at once. It is crucial for the Usage page. */}
-            <QueryParamProvider adapter={NextAdapterPages} options={{ enableBatching: true }}>
+            <QueryParamProvider adapter={NextAdapterPages}>
               <Component {...pageProps} />
             </QueryParamProvider>
           </ChakraThemeProvider>
--- a/app/src/pages/p/[projectSlug]/usage/index.tsx
+++ b/app/src/pages/p/[projectSlug]/usage/index.tsx
@@ -34,9 +34,7 @@ import AppShell from "~/components/nav/AppShell";
 import { modelInfo } from "~/server/fineTuningProviders/supportedModels";
 import dayjs, { formatToUTCDayMonth } from "~/utils/dayjs";
 import { useSelectedProject, useStats } from "~/utils/hooks";
-
-const numberWithDefault = (num: number | string | bigint | null, defaultValue = 0) =>
-  Number(num ?? defaultValue);
+import { numberWithDefault } from "~/utils/utils";
 
 export default function Usage() {
   const { data: selectedProject } = useSelectedProject();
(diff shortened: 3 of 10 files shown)


app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
+            <TabList>
+              <Tab>
+                <Heading size="md">Cost</Heading>
+              </Tab>
+              <Tab>
+                <Heading size="md">Activity</Heading>
+              </Tab>
+            </TabList>
+            <Spacer />
+            <Card w={"180px"} colorScheme="white" color={"gray.900"} padding={"5px"}>
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: It's a nit but stylistically I would prefer omitting the `{` and `}` wrapper when the wrapped value is just a normal string. So `w="180px"`, etc.

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
+                  </Card>
+                  <Card>
+                    <CardBody>
+                      <Stat>
+                        <HStack>
+                          <StatLabel flex={1}>Training Spend</StatLabel>
+                          <Icon as={DollarSign} boxSize={4} color="gray.500" />
+                        </HStack>
+                        <StatNumber color="gray.600" fontSize={"xl"}>
+                          ${parseFloat(totalTrainingSpend.toString()).toFixed(2)}
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: Why convert this to string and then parse? Is the type of `totalTrainingSpend` `string | number` or something?

app/src/server/api/routers/usage.router.ts
```diff
@@ -65,6 +68,8 @@ export const usageRouter = createTRPCRouter({
               )
               .as("inferenceCost"),
           ])
+          .where(sql`"ul"."createdAt"`, ">=", input.startDate)
```
corbt: Is adding the conditions here necessary? Don't these queries just extend `baseQuery`, where you already defined these conditions?

app/src/server/api/routers/usage.router.ts
```diff
@@ -73,42 +78,39 @@ export const usageRouter = createTRPCRouter({
...
+      const endDate = dayjs(input.endDate);
+
+      const allDates: Date[] = [];
+      for (
+        let date = startDate;
+        date.isBefore(endDate) || date.isSame(endDate);
+        date = date.add(1, "day")
       ) {
-        dayToMatch = dayjs(periods[originalDataIndex]?.period);
+        allDates.push(date.toDate());
```
corbt, line 111: Curious what happens if we skip this? I'd have assumed that even if the data is sparse (ie we only include entries for the days when there was actual usage) that the graphing library could handle that properly by just not showing any output on that day. But totally could be wrong about that!

app/src/utils/hooks.ts
```diff
@@ -220,6 +221,31 @@ export const useTestingEntries = (refetchInterval?: number) => {
   return useStableData(result);
 };
 
+export const useStats = (
```
corbt, line 224: I'm a little confused about the behavior here. It looks like if a user sets custom dates, we define the range we show as the first day of the month of the start date to the last day of the month of the end date? That seems like surprising/unexpected behavior to me -- why not just define the range as the beginning of the day of the start date to the end of the day at the end date?

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
+  const [endDate, setEndDate] = useQueryParam("end", StringParam);
+
+  const startOfThisMonth = dayjs().startOf("month").format("YYYY-MM-DD");
+  const endOfThisMonth = dayjs().endOf("month").format("YYYY-MM-DD");
+
+  const { projectSlug } = router.query;
+
+  useEffect(() => {
+    // Prevents `href` Interpolation error in nextjs
+    if (projectSlug) {
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: What does this check do? Or alternatively, under what circumstances would `projectSlug` be false? Can we accomplish the same thing with `selectedProject.id`?

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
+                  color={"gray.900"}
+                />
+                <Text onClick={handlePrevMonthChange} cursor={"pointer"}>
+                  {formatToUTCDayMonth(startDate || startOfThisMonth)}
+                </Text>
+                <Text paddingX={"5px"}> - </Text>
+                <Text onClick={handleNextMonthChange} cursor={"pointer"}>
+                  {formatToUTCDayMonth(endDate || endOfThisMonth)}
+                </Text>
+                <Icon
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: Would it be possible to use a predefined interactive component like [IconButton](https://chakra-ui.com/docs/components/icon-button/usage) with the `disabled` prop for the behavior we want? Defining our own input behavior like this is a bit fraught, because it's easy to forget about minor UX niceties that are noticeable to the end user -- eg. in this case the cursor remains as pointer even if the button is disabled, which sends the user a contradictory signal.

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (corbt) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: Additionally, by using the `withDefault` helper from `useQueryParams` you can probably get rid of this `useEffect` entirely by just setting the default values of the `startDate`/`endDate`? I think it's actually better to show the current month by default when you load the page without that being explicit in the URL, and only update the URL query string if you navigate to non-default dates.

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
 export default function Usage() {
   const { data: selectedProject } = useSelectedProject();
 
-  const stats = api.usage.stats.useQuery(
-    { projectId: selectedProject?.id ?? "" },
-    { enabled: !!selectedProject },
-  );
+  const router = useRouter();
+
+  const [startDate, setStartDate] = useQueryParam("start", StringParam);
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: Since these are generally updated at the same time, you'd likely find it more convenient to use the `useQueryParams` hook from this library, which lets you set multiple pieces of state simultaneously. That also would let you avoid changing the batching behavior (which is probably fine to change, but unnecessary in this case).

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
-            </VStack>
           </HStack>
+
+          <TabPanels>
+            <TabPanel paddingX={0}>
+              <HStack gap={4} align="start">
+                <Card flex={1}>
+                  <CardBody>
+                    <CostGraph
+                      startDate={startDate || startOfThisMonth}
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: These default values for `startDate` and `endDate` should be unnecessary if we set default values for the query parameter properly, since it will always be defined.

app/src/components/dashboard/CustomTooltip.tsx
```diff
@@ -0,0 +1,39 @@
...
+  name: string;
+  color: string;
+  value: string;
+}
+
+function CustomTooltip({
+  active,
+  payload,
+  label,
+  beforeValue,
```
corbt: Nit: `valuePrefix` might be a bit clearer.

app/src/pages/_app.tsx
```diff
@@ -43,7 +43,8 @@ const MyApp: AppType<{ session: Session | null }> = ({
           <SyncAppStore />
           <Favicon />
           <ChakraThemeProvider>
-            <QueryParamProvider adapter={NextAdapterPages}>
+            {/* enableBatching allows to update 2 states at once. It is crucial for the Usage page. */}
+            <QueryParamProvider adapter={NextAdapterPages} options={{ enableBatching: true }}>
```
arcticfly: I've wished we had a way to do this before, nice find!

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: Yes, I agree, fixing it!

app/src/pages/p/[projectSlug]/usage/index.tsx
```diff
@@ -16,72 +16,240 @@ import {
...
+
+  const { totalInferenceSpend, totalTrainingSpend, totalInputTokens, totalOutputTokens } =
+    useMemo(() => {
+      const calculateTotal = <T extends Record<string, any>>(
+        dataArray: T[] | undefined,
+        key: keyof T,
+      ) => dataArray?.reduce((acc: number, cur: T) => acc + Number(cur[key]), 0) ?? 0;
+
+      return {
+        totalTrainingSpend: calculateTotal(stats.data?.periods, "trainingCost"),
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/usage/index.tsx]: I would calculate these totals on the server and pass them to the client. Then you could get rid of this useMemo. In general, it's best to do computations on the server where possible.

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (arcticfly) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: I do remember these types being weird. I'll look at it with Bo.

app/src/server/api/routers/usage.router.ts
Kovbo: It totally makes sense; moved these statements to the base query.

app/src/pages/_app.tsx
arcticfly: Batching may not be necessary for this PR, but it's good to know we can add this when building future features.

app/src/server/api/routers/usage.router.ts
Kovbo: The library cuts the graph in this case, displaying only a period that has some data.
It will work only if we have some actual data on day 1 and day 31.
It would show an empty graph without any dates for new projects without any usage.

app/src/utils/hooks.ts
Kovbo: Yes, it was my typo.
Changing:
startOf("month") -> startOf("day")
endOf("month") -> endOf("day")

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: IconButton should be good, changing it!

app/src/server/api/routers/usage.router.ts
corbt: Ah yes I see why you did that. Although I think the code could be simplified and made a bit faster by using the `domain` argument of the XAxis component and just setting the min/max dates there. I can't find good documentation on the recharts page, but it does look like it's part of the API.

https://recharts.org/en-US/api/XAxis#domain
https://github.com/recharts/recharts/issues/1809 (workarounds for dates listed there)

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: Managed to remove all of that by using useQueryParams.

app/src/components/dashboard/CustomTooltip.tsx
Kovbo: Nice, I like it! Changed.

app/src/pages/p/[projectSlug]/usage/index.tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/usage/index.tsx]: Awesome idea! Implemented server-side calculations.

corbt (approved):

PR #521, opened 2024-01-23
Invoices added

Kovbo:
The billing page is accessible via this link: "p/[slug]/billing". The UI link is hidden.

I have also added the Credits functionality to this commit:
The Usage page calculates the available credits by adding up all records in the CreditAdjustment table by the end of the selected period.
It displays credits on the Usage page and Invoice page only if a user had some credits in that period.
When a new invoice is created, we calculate the used credits and add a new negative credit adjustment to the database.
If availableCredits > totalUsage, it creates a "Paid" status invoice, with the amount of $0.

credits
invoice


app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
     @@index([projectId, createdAt, type])
 }
+
+
+model Invoice {
+    id String @id @default(uuid()) @db.Uuid
+    amount         Decimal @default(0)
+    status         InvoiceStatus @default(PENDING)
+    slug           String
+    billingPeriod  String? // To make invoices more generic. Can we charge clients for something else manually?
```
arcticfly: My initial thought is that we would add additional charges on to a monthly invoice, but I'm not married to the idea.

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+
+    projectId String  @db.Uuid
+    project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+    description Json? // Array of objects: { [key: string]: string | number }
+
+    usageLogs UsageLog[]
+
+    createdAt DateTime @default(now())
```
arcticfly: Nit: place createdAt at the bottom of the schema to match existing tables. Our typical schema structure is as follows

id
scalar fields
relations
createdAt
indexes

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+    remainingCredits Decimal?
+
+    creditAdjustments CreditAdjustment[]
+
+    @@index([projectId, createdAt])
+}
+
+model CreditAdjustment {
+    id String @id @default(uuid()) @db.Uuid
+    amount         Decimal //Decimal?
```
arcticfly: Nit: remove comment

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+    
+    projectId String  @db.Uuid
+    project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+    description String?
+
+    Invoice Invoice? @relation(fields: [invoiceId], references: [id], onDelete: SetNull)
+    invoiceId String?  @db.Uuid
```
arcticfly, line 681: I think you need a createdAt field here so that you know whether you should associate a CreditAdjustment with the current month.

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+}
+
+enum InvoiceStatus {
```
arcticfly: Nit: define InvoiceStatus just above Invoice. In most cases we define enums in the schema file just before they are used.

app/src/components/billing/BillingContentTabs/BillingContentTabs.tsx
```diff
@@ -0,0 +1,24 @@
+import ContentTabs from "~/components/ContentTabs";
+import Invoices from "./Invoices/Invoices";
+
+export const INVOICES_TAB_KEY = "invoices";
+export const PAYMENT_METHODS_TAB_KEY = "payment-methods";
```
arcticfly: I would remove this variable if we aren't using it.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useInvoices } from "~/utils/hooks";
+import InvoicesTable from "./InvoicesTable";
+import NewTestInvoiceButton from "./NewTestInvoiceButton";
+import { FaTable } from "react-icons/fa";
+
+const Invoices = () => {
+  const query = useInvoices(10000);
+
+  const pendingCount = query.data?.invoices.filter((inv) => inv.status === "PENDING").length;
+  const paidCount = query.data?.invoices.filter((inv) => inv.status !== "PENDING").length;
```
arcticfly: I would rename this variable to something like "processedCount", since a REFUNDED invoice could count toward this total.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    <Td isNumeric>
+                      <ProjectLink
+                        href={{ pathname: "/billing/invoices/[id]", query: { id: invoice.id } }}
+                      >
+                        <Text as={"b"} color="gray.900">
+                          ${parseFloat(invoice.amount.toString()).toFixed(2)}
+                        </Text>
+                      </ProjectLink>
+                    </Td>
+                    <Td w="400px" isNumeric>
```
arcticfly: Good thinking. CANCELED and REFUNDED rows should look the same.

app/src/components/datasets/DatasetContentTabs/DatasetContentTabs.tsx
```diff
@@ -15,7 +16,7 @@ const tabs = [
     component: <General />,
   },
   {
-    key: "models",
+    key: DATASET_MODELS_TAB_KEY,
```
arcticfly: Since we don't reference this variable anywhere else, I'd actually prefer not defining it. The other tab keys are explicitly defined and exported because we navigate to them from external links.

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/TableRow.tsx
```diff
@@ -135,21 +147,19 @@ export const TableRow = ({
...
 
 const EntrySplit = ({ split }: { split: string }) => {
-  const color = split === DatasetEntrySplit.TRAIN ? "orange.500" : "purple.500";
+  const color = split === DatasetEntrySplit.TRAIN ? "orange" : "purple";
   return (
     <HStack justifyContent="flex-end">
-      <Text
-        fontSize="xs"
-        fontWeight="semibold"
+      <Badge
```
arcticfly, line 153: Could you add before and after screenshots for this change?

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
```diff
@@ -0,0 +1,149 @@
...
+  const router = useRouter();
+  const id = router.query.id as string;
+
+  const { data } = useInvoice(id, 10000);
+  const invoice = data?.invoice;
+
+  if (!invoice) return <></>;
+
+  // Checks if the description is an array of objects with string values to prevent rendering errors
+  const isCorrectDescriptionFormat = (value: unknown): value is { [key: string]: string }[] => {
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: This would be a great place to use zod. I can show you a couple places in our codebase where we're currently using it to give you a handle on how its type checking functionality works.

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
```diff
@@ -0,0 +1,149 @@
...
+                <Table size="lg" width="900px">
+                  <Thead>
+                    <Tr>
+                      <Th></Th>
+                      <Th>Description</Th>
+                      <Th isNumeric>Amount</Th>
+                    </Tr>
+                  </Thead>
+                  <Tbody>
+                    {isCorrectDescriptionFormat(invoice.description) &&
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: I would probably move this display logic and its associated type checking to a separate component within the same file.

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+    description Json? // Array of objects: { [key: string]: string | number }
+
+    usageLogs UsageLog[]
+
+    createdAt DateTime @default(now())
+    paidAt DateTime?
+
+    paymentId String?    
+
+    remainingCredits Decimal?
```
corbt: Probably shouldn't be nullable?

app/src/server/api/routers/billing.router.ts
```diff
@@ -0,0 +1,147 @@
...
+import { createTRPCRouter, protectedProcedure } from "~/server/api/trpc";
+import { kysely, prisma } from "~/server/db";
+import { v4 as uuidv4 } from "uuid";
+import { requireCanViewProject } from "~/utils/accessControl";
+import { toUTC } from "~/utils/dayjs";
+import { success } from "~/utils/errorHandling/standardResponses";
+import { getStats } from "./usage.router";
+
+export const billingRouter = createTRPCRouter({
+  invoices: protectedProcedure
```
arcticfly: I would follow our current naming convention and rename this router to be invoices.router.ts. Then I would rename `invoices` to `list`, `invoice` to `get`, `createInvoice` to `create`, etc.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,58 @@
+import { Card, VStack, Text, Icon, Link as ChakraLink } from "@chakra-ui/react";
+import { useInvoices } from "~/utils/hooks";
+import InvoicesTable from "./InvoicesTable";
+import NewTestInvoiceButton from "./NewTestInvoiceButton";
+import { FaTable } from "react-icons/fa";
+
+const Invoices = () => {
+  const query = useInvoices(10000);
```
corbt: Seems like this page is unlikely to update frequently, can we just leave the default refetching behavior?

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const pendingCount = query.data?.invoices.filter((inv) => inv.status === "PENDING").length;
```
corbt: Seems like if we're pulling the pending and paid counts right here anyway we might as well pass them into the tables directly to make sure the filter conditions stay in sync? Eg. `<InvoicesTable invoices={pendingInvoices} />`

app/prisma/schema.prisma
```diff
@@ -611,18 +613,70 @@ model UsageLog {
...
+    description String?
+
+    Invoice Invoice? @relation(fields: [invoiceId], references: [id], onDelete: SetNull)
+    invoiceId String?  @db.Uuid
+}
+
+enum InvoiceStatus {
+    PENDING
+    PAID
+    CANCELED // ?
```
corbt: "CANCELLED"

app/src/server/api/routers/billing.router.ts
```diff
@@ -0,0 +1,147 @@
...
+      return { invoices };
+    }),
+  invoice: protectedProcedure
+    .input(
+      z.object({
+        invoiceId: z.string(),
+      }),
+    )
+    .query(async ({ input, ctx }) => {
+      const invoice = await prisma.invoice.findUniqueOrThrow({
```
arcticfly: Instead of `findUniqueOrThrow` (which I admit I've used more than I should), I would make a `findUnique` call here, and explicitly throw a 404 error if the invoiceId is not found. I believe findUniqueOrThrow results in a 500 error being thrown if the record is not found.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+                      >
+                        <strong>{invoice.billingPeriod}</strong>
+                        <br />
+                        {invoice.slug}
+                      </ProjectLink>
+                    </Td>
+                    <Td w="200px" textAlign="center">
+                      <Badge
+                        variant="outline"
+                        p="1px 8px"
```
corbt: Very minor point but I prefer the `px={1} py={4}` style of props.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -0,0 +1,159 @@
...
+                );
+              })}
+            </Tbody>
+          </Table>
+        </TableContainer>
+      ) : (
+        <VStack py={8}>
+          <Icon as={FaTable} boxSize={16} color="gray.300" />
+          <Text fontWeight="bold" fontSize="md">
+            No invoices yet
```
corbt: Maybe "No invoices yet. An invoice will be created at the end of each calendar month."

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const paidCount = query.data?.invoices.filter((inv) => inv.status !== "PENDING").length;
+
+  return (
+    <VStack alignItems="flex-start" w="full" px={8} spacing={4} pb={8}>
+      {pendingCount && (
+        <>
+          <Text fontSize="2xl" fontWeight="bold">
```
corbt: Let's make this an explicit [Heading](https://chakra-ui.com/docs/components/heading/usage) element

arcticfly (commented):
Strong start

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <Text fontWeight="bold" fontSize="md">
+              No invoices yet
+            </Text>
+            <Text color="gray.500" textAlign="center" w="full" p={4}>
+              <ChakraLink
+                href="https://docs.openpipe.ai/getting-started/quick-start"
+                target="_blank"
+                color="blue.600"
+              >
+                Begin
```
corbt: I like the sentiment but this feels maybe a little too over the top. Developers sometimes get suspicious if they feel like they're being sold to too hard. 😆

I don't think we need to link out to the "getting started" page from the empty invoices state anyway though -- as a user I'm not going to be confused about why "Invoices" is empty if I just signed up, and following the "getting started" tutorial will not immediately cause an invoice to show up here anyway, so adding a link to that is arguably actually *more* confusing.

app/src/server/api/routers/billing.router.ts
```diff
@@ -0,0 +1,147 @@
...
+            .where("projectId", "=", input.projectId)
+            .where("createdAt", ">=", startOfLastMonth)
+            .where("createdAt", "<=", endOfPreviousMonth)
+            .execute();
+        }
+      });
+
+      function getLastMonthPeriod(): [Date, Date] {
+        const startOfLastMonth = toUTC(new Date()).subtract(1, "month").startOf("month").toDate();
+        const endOfPreviousMonth = toUTC(new Date()).subtract(1, "month").endOf("month").toDate();
```
corbt: Very minor but should be consistent in whether we use the word `last` or `previous` here! 😁

app/src/server/api/routers/billing.router.ts
```diff
@@ -0,0 +1,147 @@
...
+        projectId: z.string(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await requireCanViewProject(input.projectId, ctx);
+
+      const [startOfLastMonth, endOfPreviousMonth] = getLastMonthPeriod();
+
+      await kysely.transaction().execute(async (tx) => {
+        const baseQuery = tx
```
corbt: For consistency I would flip this around: first I would update the `UsageLog` table to associate the correct items with the invoice, and then I would query explicitly the `UsageLog` entries that are associated with the invoice to get the stats.

Theoretically that should have the same behavior as what you have here, but I'd be much more comfortable knowing for sure that the usage logs we're using to compute the total are definitely the exact same ones we're associating with the invoice.

app/src/server/tasks/generateTestSetEntry.task.ts
```diff
@@ -166,14 +166,24 @@ export const generateTestSetEntry = defineTask<GenerateTestSetEntryJob>({
...
           await prisma.usageLog.create({
             data: {
               fineTuneId: fineTune.id,
               projectId: fineTune.projectId,
+              baseModel: fineTune.baseModel,
               type: UsageType.TESTING,
               inputTokens,
               outputTokens,
-              cost: calculateCost(fineTune, 0, inputTokens, outputTokens),
+              cost,
```
corbt: Minor point, but you could save a few lines all the places you do this by just changing it to `...calculateCost(fineTune, 0, inputTokens, outputTokens)`

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
Comment (corbt) [reply app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Yes, specifically, we should created a `typedInvoice` helper in the same way we have `typedDatasetEntry`, `typedLoggedCall`, etc.

corbt (commented):
Excited to get this in!

app/prisma/schema.prisma
Kovbo: 0 by default should work as well. I think this column is for information purposes only. Should display the "remaining credits" on the invoice if it is not 0?

app/prisma/schema.prisma
Kovbo: Got it! Nice practice, will follow this rule.

app/prisma/schema.prisma
Kovbo: Oh, sure, thank you for noticing that!

app/prisma/schema.prisma
Kovbo: Noted, will be done.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: I was thinking of the exact same name, but I thought it would be unclear.
Maybe "historyCount" will be even better.

app/src/components/datasets/DatasetContentTabs/DatasetContentTabs.tsx
Kovbo: Noted!
Will remove it.

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/TableRow.tsx
Kovbo: t_vs_b
Left - Text
Right - Badge

I thought of replacing Text with Badge because Chakra has an exact UI component that looks very similar.

arcticfly: Thanks, the screenshot is helpful. For future reference, we often include UI changes in the main description of a PR (see https://github.com/OpenPipe/OpenPipe/pull/23 for an example).

app/prisma/schema.prisma
arcticfly: I think we're actually going to use this column when calculating how many credits are still available in the current month or used in previous months. The equation for currently available credits will look something like:

creditsLeftThisMonth = lastMonthInvoice.remainingCredits + allNewCreditsThisMonth - usageThisMonth.

app/src/server/api/routers/billing.router.ts
Kovbo: Got it, I will follow this convention!

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: In this case, I'm not sure; I personally like to have the whole table in a single component if it is not that big, or at least extract rows into separate components. But I can easily do it to see how it looks.

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Awesome!
I would like to learn more about it.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: Noted! Will be done.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: It is a good point. I was thinking of doing it but moved the simplest way to avoid typing issues.
Once I have a `typedInvoice` helper it would be much easier to do.

app/prisma/schema.prisma
Kovbo: Noted! Fixing.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
Kovbo: Noted! `px={2} py={0.5}` look similar, changed it.

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: Yes, replaced!

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: I agree; changed it to: "No invoices yet. An invoice will be created at the end of each calendar month."

app/src/server/api/routers/billing.router.ts
Kovbo: Oh yes, it looks weird, fixing it :)

app/src/server/api/routers/billing.router.ts
Kovbo: It adds another step in this transaction, but it is a valid point. Changed it.

app/src/server/tasks/generateTestSetEntry.task.ts
Kovbo: Nice hint!
I don't know which option I prefer more since it hides the actual variables, but it definitely looks succinct. Replaced it in some places.

app/src/components/datasets/DatasetContentTabs/General/DatasetEntriesTable/TableRow.tsx
Kovbo: Got it! Will include screenshots in the next PR.

app/prisma/schema.prisma
Kovbo: Got it. I thought we would use the creditAdjustment table (as a kind of event sourcing) to calculate the remaining credits.

app/prisma/schema.prisma
Kovbo: The simplest solution would be adding additional charges to the description.

app/prisma/schema.prisma
arcticfly: We'll do both. That's where allNewCreditsThisMonth comes into play.

Kovbo: It seems like we have 2 sources of truth in this case?

I was thinking of using only the creditsAjustments table, like:
`creditsLeftThisMonth = sum(creditsAjustments) - usageThisMonth.`
Did I understand it wrong?

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
```diff
@@ -0,0 +1,50 @@
+import { Card, VStack, Text, Icon, Heading } from "@chakra-ui/react";
+import { useInvoices } from "~/utils/hooks";
+import InvoicesTable from "./InvoicesTable";
+import NewTestInvoiceButton from "./NewTestInvoiceButton";
+import { FaTable } from "react-icons/fa";
+
+const Invoices = () => {
+  const query = useInvoices();
```
arcticfly, line 8: Nit: since you're only using data from the query, I would rewrite this as something like:

```
invoices = useInvoices().data;
```

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
```diff
@@ -0,0 +1,173 @@
...
+  const id = router.query.id as string;
+
+  const query = useInvoice(id);
+  const invoice = query.data;
+
+  if (!query.isLoading && !invoice) {
+    return (
+      <AppShell title="Dataset not found">
+        <Center h="100%">
+          <Text>Dataset not found 😕</Text>
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Change to "Invoice not found"

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const router = useRouter();
+  const id = router.query.id as string;
+
+  const query = useInvoice(id);
+  const invoice = query.data;
+
+  if (!query.isLoading && !invoice) {
+    return (
+      <AppShell title="Dataset not found">
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Change

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
```diff
@@ -0,0 +1,173 @@
...
+              </HStack>
+            </Button>
+          </HStack>
+        </VStack>
+      </VStack>
+    </AppShell>
+  );
+}
+
+function InvoiceBody(prop: { description: JsonValue }) {
```
Comment (arcticfly) [inline app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Nit: `prop` -> `props`

app/src/components/billing/BillingContentTabs/Invoices/Invoices.tsx
Kovbo: I use it to call `query.isLoading` to show "No invoices yet" only if the page is loaded.

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Oh, my mistake, fixing it.

app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/billing/invoices/[id].tsx]: Changed

app/prisma/schema.prisma
```diff
@@ -613,22 +615,70 @@ model UsageLog {
...
+    amount         Decimal @default(0)
+    status         InvoiceStatus @default(PENDING)
+    slug           String
+    billingPeriod  String? // MM YYYY
+    paidAt         DateTime?
+
+    description Json? // Array of objects: { [key: string]: string }
+    paymentId String?
+
+    remainingCredits Decimal?
```
arcticfly: Did we decide that we no longer need this field?

arcticfly (commented):
Once your migration is updated, I think it's ready to merge in.

app/prisma/schema.prisma
```diff
@@ -613,22 +615,76 @@ model UsageLog {
...
+
+    projectId String  @db.Uuid
+    project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
+
+    createdAt DateTime @default(now())
+
+    @@index([projectId, createdAt])
+}
+
+enum CreditAdjustmentType {
```
Kovbo, line 666: In the latest commit, I added this CreditAdjustmentType enum to the schema. What do you think @corbt , do we need it?

app/prisma/schema.prisma
Kovbo: Removed it in the latest commit.

app/prisma/schema.prisma
Kovbo: We could remove `type` and use only `invoiceId` on `CreditAdjustment` table (we already have this relation) to associate negative credit adjustments with an invoice.
CreditAdjustmentType may be useful if we have different types of credit adjustments, and if we want to explicitly track them.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <Card width="100%" overflowX="auto">
+      <TableContainer>
+        <Table variant="simple">
+          <Thead>
+            <Tr>
+              <Th>ID</Th>
+              <Th textAlign="center">Status</Th>
+              <Th isNumeric>Amount</Th>
+              {/* <Th isNumeric></Th> */}
```
arcticfly: Remove

arcticfly (approved):
Great work!

app/prisma/schema.prisma
corbt: Apologies for the belated review. Yes, I believe we can do without this field based on the way the `CreditAdjustment` table works now.

PR #524, opened 2024-01-23
Don't try to update the launch config unnecessarily

corbt:
Every time we deploy Pulumi wants to update our k8s cluster to the latest version of the AMI base image. This is unnecessary and leads to downtime because for some reason Pulumi's rollout strategy doesn't bring up the new nodes until after the old ones are gone. There's no reason to always be on the latest AMI, so let's just leave it at a version we know works.

Also removes graphana/prometheus from the cluster for now because we aren't using it for anything atm anyway.


arcticfly (approved):
Thanks, big QoL improvement

PR #525, opened 2024-01-23
Can only export OpenPipe weights

corbt:
We can't actually export your OpenAI weights, sorry!


arcticfly (approved):

arcticfly:
Talk about a competitive advantage though.

PR #526, opened 2024-01-24
Testing Anyscale

corbt:
This code deploys an Anyscale cluster for Mistral FT Optimized. It mirrors all traffic from that model to the Anyscale cluster. Currently, it then throws away the completion we get back entirely -- this is just for initial load testing/error tracking. If all looks well we can start actually using the completions in a follow-up.


arcticfly (approved):
Does Anyscale read the base models and LoRAs from the same buckets as Modal?

corbt:
> Does Anyscale read the base models and LoRAs from the same buckets as Modal?

Not exactly -- modal actually reads them from the shared modal fs directly, and anyscale reads them from s3.

PR #533, opened 2024-01-30
Stripe payments

Kovbo:
Do not merge!

**STRIPE SETUP**

Add to .env:
```
STRIPE_PUBLISHABLE_KEY
STRIPE_SECRET_KEY
STRIPE_WEBHOOK_SECRET
```

Setup a new webhook: 
Visit: https://dashboard.stripe.com/webhooks 
Add this API endpoint:  
https://app.openpipe.ai/api/stripe/webhook

Handled events:
```
charge.succeeded
charge.failed
charge.expired
```
payment_intent.succeeded 

We can probably add more events in the stripe dashboard and check for events that we may be interested in. 

We need to create temps of use. Stripe requires:
_Add terms to your website or app that state how you plan to save payment method details and allow customers to opt in. If you want to charge them when they’re offline, make sure your terms include the following:
•	The customer’s agreement to your initiating a payment or a series of payments on their behalf for specified transactions.
•	The anticipated timing and frequency of payments (for example, if the charges are for scheduled installments, subscription payments, or unscheduled top-ups).
•	How you determine the payment amount.
•	Your cancellation policy, if the payment method is for a subscription service.
Make sure you keep a record of your customer’s written agreement to these terms_.

**CRON Jobs**
There are 2 cron jobs:
`generateInvoices.task.ts` - runs at the beginning of each month and creates invoices.
`chargeInvoices.task.ts` - runs once a day. An open question: should we track failed payment attempts and stop trying to charge an account at some point? We could run the job once a month to limit the amount of attempts.

**DB changes**
Added 2 columns:
`project.billable` and `project.stripeCustomerId`

stripeCustomerId connects a stripe user to a project. Billable defines if we issue invoices for a particular project.
We will need to manually set billable=false for some clients.

The current logic for the creation of invoices is:
```
 if (!invoiceAlreadyExists && project.billable && project.id === "89149c5d-aeda-49f7-b668-41104aef8444") {
  await createInvoice(project.id, startOfPreviousMonth, endOfPreviousMonth);
}
```

We can test the creation of invoices and off-site payments on a single project: "SDK test Project".
To backfill invoices we can temporarily change "startOfPreviousMonth" to "startOfPreviousYear".

"Credits backfill"

New projects will automatically receive 100 credits.
To backfill existing projects, use this script: `backfillCredits.ts`
It adds 100 credits to all the existing projects.


app/src/components/billing/BillingContentTabs/PaymentMethods/AddPaymentMethodButton.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </HStack>
+      </Button>
+      <PaymentDetailsModal clientSecret={clientSecret} disclosure={disclosure} />
+    </>
+  );
+}
+
+// Call `loadStripe` outside of a component’s render to avoid recreating the `Stripe` object on every render.
+const stripePromise = loadStripe(
+  "pk_test_51OcEiyIpUtvR6wbgpoMtjp7GzrWoNcjM2kLSeYlEAcP9BevVtv69TeUvhndrg87A4zigWNXYfTjeHyDqX4dt3Pm100pZ8BTBtu",
```
arcticfly: Should this go in our environment variables?

app/src/components/billing/BillingContentTabs/PaymentMethods/AddPaymentMethodButton.tsx
```diff
@@ -0,0 +1,221 @@
...
+      });
+    } else {
+      const { setupIntent } = await stripe.retrieveSetupIntent(clientSecret);
+
+      return setupIntent;
+    }
+  }
+
+  return (
+    <div style={{ width: "100%" }}>
```
arcticfly: To stay consistent with Chakra styling, use `Box` here instead.

app/src/components/billing/BillingContentTabs/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const PaymentMethods = () => {
+  const selectedProject = useSelectedProject().data;
+  const paymentMethods = usePaymentMethods().data;
+  const utils = api.useContext();
+
+  const setDefaultPaymentMethodMutation = api.payments.setDefaultPaymentMethod.useMutation();
+  const deletePaymentMethodMutation = api.payments.deletePaymentMethod.useMutation();
+
+  const [setDefaultPaymentMethod, isDefaultChangeLoading] = useHandledAsyncCallback(
```
arcticfly: Nit: rename `isDefaultChangeLoading` to something like `setDefaultInProgress`

app/src/components/billing/BillingContentTabs/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,114 @@
...
+        projectId: selectedProject.id,
+        paymentMethodId,
+      });
+
+      await utils.payments.getPaymentMethods.invalidate();
+    },
+    [selectedProject, utils, setDefaultPaymentMethodMutation],
+  );
+
+  const [deletePaymentMethod, isDeleteLoading] = useHandledAsyncCallback(
```
arcticfly: Nit: Change `isDeleteLoading` to something like `deletionInProgress`

app/src/components/billing/BillingContentTabs/PaymentMethods/PaymentMethods.tsx
```diff
@@ -0,0 +1,114 @@
...
+                  ) : (
+                    <Button
+                      onClick={() => {
+                        if (!isDefaultChangeLoading) setDefaultPaymentMethod(method.id);
+                      }}
+                      fontSize="sm"
+                      px={1}
+                      variant="ghost"
+                      _hover={{ bgColor: "transperent" }}
+                    >
```
arcticfly, line 87: Instead of adding custom <Spinner /> elements in loading buttons, you can use Chakra's built-in `isLoading` prop. That should also disable the button while it's loading.

app/src/pages/api/stripe/webhook.ts
```diff
@@ -0,0 +1,124 @@
...
+  try {
+    event = stripe.webhooks.constructEvent(rawBody, sig ?? "", endpointSecret);
+  } catch {
+    return res.status(400).send(`Webhook Error`);
+  }
+
+  switch (event.type) {
+    case "charge.succeeded":
+      const paymentIntent = event.data.object;
+      const invoiceId = paymentIntent.metadata.invoiceId;
```
arcticfly, line 48: This seems like a good place to check if the invoiceId exists. That way we can change the function signature of all the functions you call later to expect that invoiceId is a string, not string or undefined.

app/src/pages/api/stripe/webhook.ts
```diff
@@ -0,0 +1,124 @@
...
+      break;
+  }
+
+  res.json({ received: true });
+};
+
+export default handler;
+
+async function notifyAdminsAboutFailure(invoiceId: string | undefined) {
+  const project = await findProject(invoiceId);
```
arcticfly: Since we only use findProject in one place, I don't think we need to define it as a separate function (unless you plan to use it in more places later?)

Also, I think we should capture an exception in Sentry if we aren't able to find an invoice or a project. Search the codebase for `captureException` to see how we do that.

arcticfly (commented):
I haven't fully reviewed, but I want to avoid conflicts with @corbt 😆

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -96,15 +129,20 @@ const InvoicesTable = ({ invoices }: Props) => {
...
                         variant="ghost"
                         color="blue.500"
-                        onClick={(e) => e.stopPropagation()}
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          pay(invoice.id);
+                        }}
                       >
-                        Pay
+                        {isPaymentLoading && processingInvoceId === invoice.id ? (
```
corbt: There's actually an `isLoading` prop on buttons that handles this. I usually also set `isDisabled` to true when it's processing to prevent someone from accidentally double-clicking the button.

app/src/components/billing/BillingContentTabs/PaymentMethods/AddPaymentMethodButton.tsx
Kovbo: Oh, I was testing and forgot to replace it!

app/src/server/utils/stripe.ts
```diff
@@ -0,0 +1,144 @@
...
+  return customer;
+}
+
+// Overload signatures
+export function getStripeCustomerId(input: { projectId: string; required: true }): Promise<string>;
+export function getStripeCustomerId(input: {
+  projectId: string;
+  required: false;
+}): Promise<string | null>;
+export async function getStripeCustomerId(input: { projectId: string; required: boolean }) {
```
arcticfly, line 126: Good use of overloaded function signatures.

app/src/server/utils/stripe.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+export async function createStripeCustomerAndConnectItToProject(projectId: string) {
+  // Find a project
+  const project = await prisma.project.findUnique({
+    where: {
+      id: projectId,
+    },
+  });
+
```
arcticfly, line 98: I would check to ensure that project exists here and throw a 404 if it doesn't. That way you don't need all the `project?.` later in the function, and we don't create customers for nonexistent projects.

app/src/server/utils/stripe.ts
```diff
@@ -0,0 +1,144 @@
...
+  // Create stripe customer
+  const customer = await stripe.customers.create({
+    name: `${project?.name ?? "Project"} - ${project?.slug ?? "OpenPipe"}`,
+    metadata: {
+      projectId: projectId,
+    },
+  });
+
+  //Associate customer with a project
```
arcticfly: Nit: add space between // and Associate

app/src/server/emails/sendPaymentFailed.ts
```diff
@@ -0,0 +1,25 @@
...
+  projectSlug,
+}: {
+  recipientEmail: string;
+  projectName: string;
+  projectSlug: string;
+}) => {
+  const paymentMethodsLink = `${env.NEXT_PUBLIC_HOST}/p/${projectSlug}/billing/payment-methods`;
+
+  const emailBody = `
+  <p>The payment for ${projectName} has failed. This could be due to an issue with your payment method or other transaction-related problems.</p>
```
corbt: Let's replace the first sentence with the following: "We attempted to charge your default payment method for project ${projectName} but the payment failed."

app/src/server/emails/sendPaymentFailed.ts
```diff
@@ -0,0 +1,25 @@
+import { env } from "~/env.mjs";
```
corbt, line 1: We should also have an email draft for when the payment went through successfully.

app/src/server/api/routers/payments.router.ts
```diff
@@ -0,0 +1,173 @@
...
+    )
+    .mutation(async ({ input, ctx }) => {
+      await requireIsProjectAdmin(input.projectId, ctx);
+
+      let stripeCustomerId = await getStripeCustomerId({
+        projectId: input.projectId,
+        required: true,
+      });
+
+      // Check if stripe user was deleted. If yes - create a new one
```
arcticfly, line 49: Under what circumstances will the stripe user be deleted?

app/src/server/api/routers/payments.router.ts
```diff
@@ -0,0 +1,173 @@
...
+  getStripeCustomerId,
+  setDefaultPaymentMethod,
+  createSetupIntent,
+  getStripeCustomer,
+  deletePaymentMethod,
+} from "~/server/utils/stripe";
+import { chargeInvoice } from "~/server/tasks/chargeInvoices.task";
+
+export const paymentsRouter = createTRPCRouter({
+  createStripeCustomer: protectedProcedure
```
corbt: I don't fully grok the flow yet, but will `createStripeCustomer` and `createStripeIntent` always be called at the same time? If so, should they be combined into a single mutation?

app/src/server/utils/stripe.ts
```diff
@@ -0,0 +1,144 @@
...
+    payment_method: paymentMethodId,
+    return_url: returnUrl,
+    off_session: true,
+    confirm: true,
+    metadata: {
+      invoiceId, // Used to track status using webhooks
+    },
+  });
+}
+export async function createSetupIntent(stripeCustomerId: string) {
```
arcticfly: You could make these functions even more streamlined with arrow function syntax:

```
export const createSeupIntent = async (stripeCustomerId: string) =>
stripe.setupIntents.create({
...
```

app/src/components/billing/BillingContentTabs/PaymentMethods/AddPaymentMethodButton.tsx
Kovbo: Yes, fixed!

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  try {
+    const paymentIntent = await createStripePaymentIntent({
+      amount: usdToCents(invoice.amount),
+      invoiceId: invoiceId,
+      stripeCustomerId: project.stripeCustomerId,
+      paymentMethodId: defaultPaymentMethodId,
+      returnUrl: `${
+        env.NEXT_PUBLIC_HOST ?? "app.openpipe.com"
+      }/p/${project?.slug}/billing/inboices`,
```
corbt: Does this link work? It looks like "invoices" has a typo at a minimum.

app/src/components/billing/BillingContentTabs/PaymentMethods/PaymentMethods.tsx
Kovbo: So convenient, I did not know it! Changed in all the places.

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+          paidAt: new Date(),
+          status: "PAID",
+          paymentId: paymentIntent.id,
+        },
+      });
+
+      return success("Payment is succeeded!");
+    }
+
+    return error("Peyment is processing");
```
corbt: "Payment" spelling

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+          id: invoiceId,
+        },
+        data: {
+          paidAt: new Date(),
+          status: "PAID",
+          paymentId: paymentIntent.id,
+        },
+      });
+
+      return success("Payment is succeeded!");
```
corbt: "Payment successfully applied"

app/src/server/api/routers/payments.router.ts
```diff
@@ -0,0 +1,173 @@
...
+        const paymentMethods = await getPaymentMethods(stripeCustomerId);
+
+        if (!paymentMethods) {
+          return null;
+        }
+
+        // Set initial payment method as default
+        let defaultPaymentMethodId = await getDefaultPaymentMethodId(stripeCustomerId);
+        const availablePaymentMethodId = paymentMethods.data[0]?.id ?? undefined;
+        if (defaultPaymentMethodId === undefined && availablePaymentMethodId) {
```
arcticfly: Nit: `defaultPaymentMethodId === undefined` -> `!defaultPaymentMethodId`

app/src/server/api/routers/payments.router.ts
```diff
@@ -0,0 +1,173 @@
...
+      await requireIsProjectAdmin(input.projectId, ctx);
+
+      const stripeCustomerId = await getStripeCustomerId({
+        projectId: input.projectId,
+        required: true,
+      });
+
+      try {
+        const response = setDefaultPaymentMethod(stripeCustomerId, input.paymentMethodId);
+        console.log(response);
```
arcticfly: Remove

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+});
+
+export async function chargeInvoice(invoiceId: string) {
+  const invoice = await prisma.invoice.findUnique({
+    where: {
+      id: invoiceId,
+    },
+  });
+
+  if (invoice?.status !== "PENDING" || Number(invoice.amount) < 1) {
```
arcticfly, line 37: If the invoice status is ERROR (due to an earlier payment method being declined) shouldn't we let the user pay in this function?

arcticfly: I'd also prefer not to check the amount of the invoice here. If the invoice is below a certain threshold, we should just mark it as cancelled when we create it.

app/src/pages/api/stripe/webhook.ts
Kovbo: For some reason, I could not do it. returning and breaking the switch/case did not help with typings. Let me take a look again.

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+    where: {
+      id: invoice.projectId,
+    },
+  });
+
+  if (!project || !project.stripeCustomerId) {
+    return error("Add a default payment method.");
+  }
+
+  let defaultPaymentMethodId = await getDefaultPaymentMethodId(project.stripeCustomerId);
```
arcticfly: Since we may actually set this variable to a non-default payment method, let's rename it to something like `paymentMethodToUse`

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  if (!defaultPaymentMethodId) {
+    // Search for available payment methods and use the first one
+    const paymentMethods = await getPaymentMethods(project.stripeCustomerId);
+
+    if (paymentMethods && paymentMethods.data[0]?.id) {
+      defaultPaymentMethodId = paymentMethods.data[0]?.id;
+    } else {
+      return error("Add a default payment method.");
```
arcticfly: Shouldn't this error be "Add a payment method."?

app/src/pages/api/stripe/webhook.ts
Kovbo: We might need to add other event handlers and send notifications about successful payments. We would need a separate function in this case, but I will remove it for now.

Added `captureException,` but I am unsure if I did it correctly, since the codebase has different implementations.

app/src/server/tasks/chargeInvoices.task.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  try {
+    const paymentIntent = await createStripePaymentIntent({
+      amount: usdToCents(invoice.amount),
+      invoiceId: invoiceId,
+      stripeCustomerId: project.stripeCustomerId,
+      paymentMethodId: defaultPaymentMethodId,
+      returnUrl: `${
+        env.NEXT_PUBLIC_HOST ?? "app.openpipe.com"
```
arcticfly: Do you have to include this `?? "app.openpipe.com"` logic? I think we require `NEXT_PUBLIC_HOST` to be set.

app/src/server/tasks/generateInvoices.task.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export const generateInvoices = defineTask({
+  id: "generateInvoices",
+  handler: async () => {
+    const [startOfPreviousMonth, endOfPreviousMonth] = getPreviousMonthPeriodUTC();
+
+    const projects = await prisma.project.findMany();
+
+    for (const project of projects) {
+      //Avoid creating multiple invoices for the same billing period
```
arcticfly: Nit: add space between // and the rest of comments to improve readability.

app/src/server/api/routers/projects.router.ts
```diff
@@ -258,6 +258,14 @@ export const projectsRouter = createTRPCRouter({
             lastViewedProjectId: newProjectId,
           },
         }),
+        prisma.creditAdjustment.create({
```
corbt: Actually let's only do this for the first default project we create when a user signs up, instead of for every new project an existing user creates.

app/src/server/tasks/generateInvoices.task.ts
```diff
@@ -0,0 +1,180 @@
...
+              type: "INVOICE",
+            })
+            .execute();
+        }
+      }
+    }
+  });
+}
+
+function getInvoiceSlug(length = 6) {
```
arcticfly: Nit: I would use "generate" instead of "get" in the names of these functions. "get" implies that we're retrieving information that already exists, whereas "generate" implies that we're creating a new value.

app/src/server/tasks/worker.ts
```diff
@@ -63,6 +67,18 @@ const runner = await run({
...
     },
+    {
+      task: generateInvoices.task.identifier,
+      // run at the beginning of each month
+      pattern: "0 0 1 * *",
+      identifier: generateInvoices.task.identifier,
+    },
+    {
+      task: chargeInvoices.task.identifier,
+      // run once a day
```
arcticfly: I think we should only run this once a month to avoid overdraft fees on our users' payment methods if they don't have sufficient funds.

app/src/server/utils/emails.ts
arcticfly, line 1: I would move this file into src/server/emails and rename if something specific to its task like `emailAdminsAboutPaymentFailure.ts`.

arcticfly (commented):
Nice. You've already thought through a variety of tricky edge cases.

corbt (commented):
Looks pretty good overall!

app/src/server/utils/stripe.ts
Kovbo: Completely agree; changed it.

app/src/server/api/routers/payments.router.ts
Kovbo: A user can be deleted through the stripe dashboard. In this case, it would throw an error.
A stripe user has a type of `Stripe.Response<Stripe.Customer | Stripe.DeletedCustomer>`.

app/src/server/utils/stripe.ts
Kovbo: Yes, it is a good place for an arrow function.

app/src/server/tasks/chargeInvoices.task.ts
Kovbo: Currently, an invoice would have a "PENDING" status if payment was declined. This way, the user can update payment details and pay it manually, or the system will charge automatically once the corn job executes.

When creating an invoice, I cancel it by default if the amount is <= 1. We can probably remove this check, but since the payment logic is different from the invoice creation logic, I explicitly added it.

app/src/server/tasks/chargeInvoices.task.ts
Kovbo: Nice idea, changed.

app/src/server/tasks/chargeInvoices.task.ts
Kovbo: Yes, good idea.

app/src/server/tasks/chargeInvoices.task.ts
Kovbo: Oh yes, I wrote it before I discovered the `env.mjs` helper :)

app/src/server/tasks/generateInvoices.task.ts
Kovbo: I see, nice idea!

app/src/server/tasks/worker.ts
Kovbo: Yes, it may be a solution since we have a manual "pay" button.
I guess we can create invoices on the 1st day of the month and charge clients on the second?

app/src/server/utils/emails.ts
Kovbo: Ok, got it! I considered the emails folder as a collection of email templates. I changed the approach a bit. I created a function `sendToAdmins` that accepts a specific email template as a callback.

app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
Kovbo: Oh yeah, I just discovered it! This way, the code looks much better.

app/src/server/emails/sendPaymentFailed.ts
Kovbo: Okay, will surely do it!

app/src/server/emails/sendPaymentFailed.ts
Kovbo: Will take a look at it.

app/src/server/tasks/chargeInvoices.task.ts
Kovbo: Oh, yes, sorry. Actually the `returnUrl` is redundant here as we use off-site payments without redirections, but for some reason stripe includes it in docs.

app/src/server/emails/sendPaymentFailed.ts
Kovbo: Added payment successful emails.

app/src/server/api/routers/payments.router.ts
Kovbo: Yes, it should work fine!

app/src/server/api/routers/projects.router.ts
Kovbo: Sure, changed it!

app/src/server/tasks/worker.ts
arcticfly: Ideally not too much time would pass between the creation of invoices and the time we charge them. Could we create the invoices at midnight, then charge them a few hours later, perhaps around 8am UTC?

app/src/server/utils/emails.ts
arcticfly: I don't currently see that function in this PR. Have you just not pushed it yet?

corbt (approved):
Ok, I think this is good to merge once the merge conflicts are fixed. Nice work!

app/src/pages/api/stripe/webhook.ts
arcticfly: Ah, I think I see the problem. You need to define invoiceId before the switch statement instead of in one of its conditional blocks.

app/src/server/utils/emails.ts
arcticfly: My mistake, I was looking at an old version of the code.

app/src/server/tasks/worker.ts
Kovbo: Good idea, I will change it once we test the production payments!

PR #534, opened 2024-01-30
More-anyscale

corbt:
Mirror production traffic to a cluster of A10s to benchmark their perf vs A100s.


app/scripts/copy-model-weights.ts
```diff
@@ -28,8 +28,10 @@ if (model.pipelineVersion < 3) {
 await $$`mkdir -p /tmp/${modelSlug}`;
 await $$`rm -rf /tmp/${modelSlug}/*`;
 
-console.log("Transferring the s3 weights...");
-await $$`aws s3 sync s3://user-models-pl-prod-5e7392e/models/${model.baseModel}:${model.id}:1 s3://user-models-pl-stage-4c769c7/models/${model.baseModel}:${model.id}:1`;
+const outPath = `s3://user-models-pl-stage-4c769c7/models/${model.baseModel}:${model.id}:1`;
+
+console.log(`Transferring the s3 weights to ${outPath}...`);
+await $$`aws s3 sync s3://user-models-pl-prod-5e7392e/models/${model.baseModel}:${model.id}:1 ${outPath}`;
```
arcticfly, line 34: I guess these paths are unlikely to change, but I would still feel better if they were imported from the `buckets` variable in deploy.ts

arcticfly (approved):

app/scripts/copy-model-weights.ts
corbt: Yeah we use them in a number of places; will centralize at some point.

PR #537, opened 2024-01-31
Stripe env variable added

Kovbo:
Public env variable added


corbt (approved):

infra/src/app-image.ts
```diff
@@ -12,6 +12,7 @@ export const appImage = new awsx.ecr.Image(nm("app"), {
     NEXT_PUBLIC_HOST: appUrl,
     NEXT_PUBLIC_SENTRY_DSN: getConfig("NEXT_PUBLIC_SENTRY_DSN"),
     NEXT_PUBLIC_DEPLOY_ENV: getConfig("NEXT_PUBLIC_DEPLOY_ENV"),
+    NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY: getConfig("NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY"),
```
corbt, line 15: I think you need to rename this in `Pulumi.prod.yaml` as well?

corbt (commented):
I think you need to rename this in `Pulumi.prod.yaml` as well?

infra/src/app-image.ts
Kovbo: Hm? I did not rename it, I added the key to these 2 files. It already has the "NEXT_PUBLIC" prefix. Or I missed something?

PR #539, opened 2024-02-01
Stripe invoces all periods

Kovbo:
Before deploying to production, turn off invoices for particular clients by manually setting the project's "billable" to false.

It runs a job that creates invoices for all the previous periods.
It will show the billable period like: "May 2023 - Jan 2024", and take into account $100 bonus credits.

The initial run may take some time since we have to calculate all the existing data. I hope it will not have any timeout errors.


corbt (approved):
Ok, I've updated billing to false for the relevant projects, this looks good to merge.

PR #540, opened 2024-02-01
Export internal Python SDK, document updating tags

arcticfly:
We're now exporting the internal SDKs from both our Python and TypeScript packages.


docs/features/updating-tags.mdx
```diff
@@ -0,0 +1,52 @@
+---
+title: "Running Inference"
```
corbt: "Updating Tags"

corbt: Actually looks like this file was copied over but retains the content on running inference?

corbt (approved):
Thanks, this looks good.

docs/features/updating-tags.mdx
arcticfly: Yep, I haven't written any of the docs yet

PR #543, opened 2024-02-01
added invoice index to usageLog and cascade credits delete

Kovbo:
It will run a job a single time, at 2 AM (3 hours from now).
It should be enough to create indexes and delete old records.


corbt (approved):
Looks good! Maybe set it earlier so the job gets started sooner and we can see the results?

corbt (approved):
Should also be fine to run this sooner so we get feedback sooner!

Kovbo:
> Looks good! Maybe set it earlier so the job gets started sooner and we can see the results?

Changed it to 2 hours from now. I don't know much till will it take to create indexes and delete all the invoices. 2 hours should be enough.

PR #544, opened 2024-02-01
Relabel properly tagged logs

arcticfly:
It's useful to relabel some request logs for retraining purposes when a fine-tuned model performs poorly on a prompt that's unlike anything it was trained on. Let's streamline the process by automatically relabeling any request log whose tags are updated to include `relabel: "true"` and `add_to_dataset: "original_model_dataset"`. We'll make this API more configurable in the future.


corbt (approved):
Very nice!

PR #547, opened 2024-02-02
Regular billing logic

Kovbo:
Set a cron job that creates invoices each month at 1 AM and charges at 8 AM.


corbt:
What emails if any will go out when we try to charge invoices given that none of them have payment methods associated with them?

Kovbo:
> What emails if any will go out when we try to charge invoices given that none of them have payment methods associated with them?

It does not send an email.
I was thinking of adding an existing "sendPaymentFailed", or creating a new email template.

arcticfly (approved):

PR #556, opened 2024-02-03
2 scripts created

Kovbo:
Call scripts manually listing ids:
pnpm tsx src/server/scripts/deleteInvoicesAndNoBill.ts 1111-11111-1111-111-1111111111
pnpm tsx src/server/scripts/zeroCheckedToZeroInvoice.ts 1111-11111-1111-111-1111111111

### DIFF
--- a/app/src/server/scripts/zeroCheckedToZeroInvoice.ts
+++ b/app/src/server/scripts/zeroCheckedToZeroInvoice.ts
@@ -1,0 +1,40 @@
+import { getPreviousMonthPeriodUTC } from "~/utils/dayjs";
+import { prisma } from "../db";
+import { createInvoice } from "../tasks/generateInvoices.task";
+
+const [startOfPreviousMonth, endOfPreviousMonth] = getPreviousMonthPeriodUTC();
+
+const projectId = process.argv[2];
+
+console.log("Doing: apply credits for all of their usage, then regenerate the invoice");
+
+const project = await prisma.project.findFirstOrThrow({
+  where: {
+    id: projectId,
+  },
+});
+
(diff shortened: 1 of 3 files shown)


corbt (approved):
Will run these, thanks!

app/src/server/scripts/zeroCheckedToZeroInvoice.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+}
+
+await prisma.invoice.deleteMany({
+  where: {
+    projectId: project.id,
+  },
+});
+
+await createInvoice(project.id, new Date(2010), endOfPreviousMonth);
```
corbt, line 40: Minor nit: can we set these as `project.createdAt` both here and in the task, instead of 2010 milliseconds past the epoch?

corbt: I'll just merge as-is so I can run this for now, can write a query to update the dates post-creation pretty easily.

app/src/server/scripts/zeroCheckedToZeroInvoice.ts
```diff
@@ -0,0 +1,40 @@
...
+});
+
+const invoice = await prisma.invoice.findFirst({
+  where: {
+    projectId: project.id,
+  },
+});
+
+if (invoice) {
+  await prisma.creditAdjustment.create({
```
corbt, line 24: I tried this in production and found that the invoices are still getting being generated for the original amount. I think the issue is that this `creditAdjustment` needs to be set to be during the invoice period, otherwise it won't be applied to the invoice.

PR #561, opened 2024-02-07
Cache project values

arcticfly:
Let's cache the number of logged calls in a project, along with the tag names for that project.


corbt (approved):
Scary. Let's do it.

PR #564, opened 2024-02-07
Invoices and emails

Kovbo:
**Added 4 emails**
inv
pr
ps
usage

We send emails if:
-payment failed
-payment succeeded
-payment not required (a user used the service, but credits covered the expenses)

To run tsx email templates outside of the next.js environment (worker or manual scripts), I added the `tsconfig.server.json` file that allows using tsx/jsx. 
The helper `"tsx": "tsx --tsconfig tsconfig.server.json"` allows to run scripts without specifying the tsconfig file, but the `"dev:worker"/"worker"` scripts still require an explicit path.

**Invoice logic change**
We are issuing invoices at 2 AM UTC on the first day of each month, and trying to charge them right away.
If a payment fails, we run a job that tries to charge again, on the next day at 8 AM. A user will have 28 hours to add a payment method. Should we change the time? Maybe add one more job that runs in a week?

**Invoices page design change**
invoice

**Database change**
Added indexes to `FineTune` model and `Dataset` model.
Added the `FREE` enum to invoice status.
Fixed a relation typo from Invoice to invoice in `CreditAdjustment` table.

**Scripts**
To rename all existing invoices from CANCELLED to FREE, use the script:
`pnpm tsx src/server/scripts/renameCancelledInvoicesToFree.ts`
I could not combine it into a single migration, postgress does not allow to do it in one transaction.

Another script sends emails to existing clients who have pending invoices or used credits.
`pnpm tsx src/server/scripts/sendInvoiceEmailToExistingClients.ts`


app/src/components/billing/BillingContentTabs/Invoices/InvoicesTable.tsx
```diff
@@ -139,6 +130,15 @@ const InvoicesTable = ({ invoices }: Props) => {
 
                     {invoice.paidAt && <Text>{dayjs(invoice.paidAt).format("MMM D, YYYY")}</Text>}
                   </Td>
+                  <Td w="200px" isNumeric>
+                    <ProjectLink
+                      href={{ pathname: "/billing/invoices/[id]", query: { id: invoice.id } }}
+                    >
+                      <Text as={"b"} color="gray.900">
```
arcticfly: Nit: as={"b"} -> as="b"

app/src/server/emails/sendInvoiceNotificationWithoutRequiredPayment.ts
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    ({ description: parsedDescription } = typedInvoice({ description }));
+  } catch (e) {
+    return;
+  }
+
+  const emailBody = render(
+    InvoiceNotificationWithoutRequiredPayment({
+      projectName,
+      amount: Number(Number(amount).toFixed(2)).toLocaleString(),
```
arcticfly, line 30: Is casting the amount to a number here necessary?

app/src/server/emails/templates/InvoiceNotificationWithoutRequiredPayment.tsx
```diff
@@ -0,0 +1,58 @@
...
+  description,
+}: Props) => {
+  const previewText = `OpenPipe Usage ${billingPeriod}`;
+
+  return (
+    <EmailLayout previewText={previewText}>
+      <Heading>OpenPipe Usage {billingPeriod}</Heading>
+
+      <Text>
+        This is your monthly <Link href={invoicesLink}>invoice</Link> for OpenPipe Project{" "}
```
arcticfly: "OpenPipe Project" -> "OpenPipe project"

app/src/server/emails/templates/InvoiceNotificationWithoutRequiredPayment.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <Text>
+        Please visit your <Link href={invoicesLink}>invoice</Link> to see more details and add a
+        payment method. This will ensure we avoid any interruption of service!
+      </Text>
+      <Button href={invoicesLink}>Details</Button>
+    </EmailLayout>
+  );
+};
+
+InvoiceNotificationWithoutRequiredPayment.PreviewProps = {
```
arcticfly, line 48: What are these PreviewProps used for?

app/src/server/emails/templates/PaymentFailed.tsx
```diff
@@ -0,0 +1,60 @@
...
+  projectLink,
+  billingPeriod,
+  description,
+}: Props) => {
+  const previewText = `Add a payment method. OpenPipe Usage ${billingPeriod}.`;
+
+  return (
+    <EmailLayout previewText={previewText}>
+      <Heading>Payment required</Heading>
+      <Text>Avoid any interruption of service by adding a payment method.</Text>
```
arcticfly: I would move this text to be just above the button to add a payment method.

arcticfly: I would also reword it to be something like: "Avoid service interruptions by adding a payment method."

app/src/server/emails/templates/ProjectInvitation.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const ProjectInvitation = ({
+  subject,
+  projectName,
+  invitationSenderName,
+  invitationSenderEmail,
+  invitationLink,
+}: Props) => {
+  const previewText = `You have been invited to join ${projectName} by ${invitationSenderName} (${invitationSenderEmail}).`;
```
arcticfly: Could we make project name bold?

app/src/server/emails/templates/ProjectInvitation.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface Props {
+  subject: string;
+  projectName: string;
+  invitationSenderName: string;
+  invitationSenderEmail: string;
+  invitationLink: string;
+}
+
+const ProjectInvitation = ({
+  subject,
```
arcticfly: Looks like we don't use this subject prop. We should remove it.

app/src/server/emails/templates/layout/Footer.tsx
```diff
@@ -0,0 +1,22 @@
+import { Row, Section, Text } from "@react-email/components";
+import React from "react";
+
+export default function Header() {
```
arcticfly: Rename to Footer

app/src/server/emails/templates/layout/Heading.tsx
```diff
@@ -0,0 +1,17 @@
+import { Text } from "@react-email/components";
+import React, { CSSProperties, ReactNode } from "react";
+
+interface Props {
+  style?: CSSProperties;
+  children: ReactNode;
+}
+export default function Header({ children, style }: Props) {
```
arcticfly: I would rename this file to be Header, since this component is more of a header than a heading.

app/src/server/emails/templates/layout/InvoiceDescription.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+export default function InvoiceDescription({ description, total, style }: Props) {
+  return (
+    <Section style={{ ...style, ...dark }}>
+      {description &&
+        description.map((item, index) => (
+          <Row key={index}>
+            <Column style={{ width: "100%" }}>
+              <Text>{item.text}:</Text>
+              {/* {item.description?.split("\n").map((i, key) => <Text key={key}>{i}</Text>)} */}
```
arcticfly: Do we need this?

app/src/server/scripts/sendInvoiceEmailToExistingClients.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+  },
+  include: {
+    invoice: true,
+    project: true,
+  },
+});
+
+for (const creditAdjustment of creditAdjustments) {
+  creditAdjustment.invoice;
```
arcticfly: Remove

app/src/server/scripts/sendInvoiceEmailToExistingClients.ts
```diff
@@ -0,0 +1,46 @@
...
+      lt: 0,
+    },
+  },
+  include: {
+    invoice: true,
+    project: true,
+  },
+});
+
+for (const creditAdjustment of creditAdjustments) {
```
arcticfly, line 20: Wouldn't it be better to cycle through all invoices?

app/src/server/tasks/generateInvoices.task.ts
```diff
@@ -19,7 +22,26 @@ export const generateInvoices = defineTask({
 
     for (const project of projects) {
       if (project.billable) {
```
arcticfly: I would add project.billable to the where clause of your findMany call.

app/src/server/tasks/generateInvoices.task.ts
```diff
@@ -29,123 +51,122 @@ export const generateInvoices = defineTask({
...
+    if (invoice && creditsUsed > 0) {
       await tx
-        .updateTable("UsageLog")
-        .set({
+        .insertInto("CreditAdjustment")
+        .values({
+          id: uuidv4(),
+          projectId: projectId,
+          amount: -creditsUsed,
+          //Adjustments should be created on the first day of the next month, when invoices are generated
```
arcticfly: Nit: add space between // and the rest of your comment.

app/src/server/tasks/worker.ts
```diff
@@ -72,17 +72,17 @@ const runner = await run({
...
     },
 
-    // TODO: Uncomment this when we are ready to charge invoices
-    // {
-    //   task: chargeInvoices.task.identifier,
-    //   // run at 8 AM UTC, on the first day of each month, after invoices are created
-    //   pattern: "0 8 1 * *",
-    //   identifier: chargeInvoices.task.identifier,
-    // },
+    //If an invoice payment failed, it will retry on the next day
```
arcticfly: Nit: add leading space

arcticfly (commented):
Our emails are going to look so much better!

app/src/server/emails/templates/InvoiceNotificationWithoutRequiredPayment.tsx
corbt: I'm guessing the react-email package uses them as the default values when previewing the template.

app/src/server/tasks/worker.ts
```diff
@@ -72,17 +72,17 @@ const runner = await run({
...
+    {
+      task: chargeInvoices.task.identifier,
```
corbt, line 81: If we're charging invoices immediately do we still need this task?

app/src/server/emails/sendInvoiceNotificationWithoutRequiredPayment.ts
Kovbo: Invoice stores amount as Decimal, so casting to a number simplifies typings.

app/src/server/emails/templates/InvoiceNotificationWithoutRequiredPayment.tsx
Kovbo: Yes, it is just a collection of placeholders for development preview.

app/src/server/emails/templates/ProjectInvitation.tsx
Kovbo: I believe most mailing services can only display a preview as text, but I can highlight it in the email body.

app/src/server/emails/templates/ProjectInvitation.tsx
Kovbo: We are using it in the `<Heading>{subject}</Heading>`

app/src/server/emails/templates/layout/Footer.tsx
Kovbo: Oops

app/src/server/scripts/sendInvoiceEmailToExistingClients.ts
Kovbo: We have a lot of invoices with $0 spent. We do not know if it had any usage, or if it is just an empty invoice since we issued invoices for all the projects, even without any usage. Cycling through the negative credit adjustments allows finding only invoices with some usage.

app/src/server/tasks/generateInvoices.task.ts
Kovbo: Good idea!

app/src/server/tasks/generateInvoices.task.ts
Kovbo: Oh, I always forget about it!

app/src/server/tasks/worker.ts
Kovbo: This task loops through pending invoices and tries to charge them again. In most cases, it means that the payment failed initially; a user received a notification and updated a payment.
We could charge each time a user updates a payment method, but there are other cases, like when a user does not have sufficient funds on a card or some banking errors. In such cases, we need to run a task that charges again.
The only question is how often we should run it.

I just added a logic that searches for pending invoices not older than 3 months. If we could not charge an invoice for 3 months, we should probably stop trying.

app/src/server/emails/templates/PaymentFailed.tsx
Kovbo: Yes, it looks good!
failed
I added a text because I feel it is not explicit enough. I also changed "Add a payment method" to "Update a payment method". In most cases, it will fail if a user already has a connected payment method.

app/src/server/emails/templates/ProjectInvitation.tsx
arcticfly: Ah I see that now. Do we need to accept the subject as a prop, or can we just hardcode it in? Does the subject message ever change?

app/src/server/scripts/sendInvoiceEmailToExistingClients.ts
arcticfly: Are we sure that there is currently only one creditAdjustment for each project?

app/src/server/emails/templates/PaymentFailed.tsx
arcticfly: Looks good! I would say "updating your payment method" instead of "updating a payment method".

app/src/server/scripts/sendInvoiceEmailToExistingClients.ts
Kovbo: Yes, each project should have only one invoice at this moment, and only one negative credit adjustment associated with it.

app/src/server/emails/templates/ProjectInvitation.tsx
Kovbo: We were using the same text in a function that sends this email. But it's okay to remove it. I do not pass subjects in other templates because the text is a bit different.

app/src/server/emails/templates/PaymentFailed.tsx
Kovbo: Changed.

corbt (approved):
Let's see how this goes!

arcticfly:
> Let's see how this goes!

Famous last words

PR #565, opened 2024-02-08
Count logged calls separately

arcticfly:
In some projects, counting logged calls takes much longer than just displaying the first 10 that match the current filters. Let's show those first matching logs before we calculate the total number.

This PR adds a splitLink that allows us to disable batching.


app/prisma/migrations/20240208002748_remove_num_logged_calls/migration.sql
```diff
@@ -0,0 +1,11 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `numLoggedCalls` on the `Project` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "Project" DROP COLUMN "numLoggedCalls";
```
corbt: I think prisma does migrations in a transaction but let's move this ALTER TABLE line to the end of the migration instead of the beginning. If it's not done in a transaction putting this at the beginning will throw errors on logged call creation until next calls finish.

corbt (approved):

PR #568, opened 2024-02-08
finetune calculator added

Kovbo:
What do you think about the UI?

cal
--UPD--

Providing some more design options:

It currently has this option:

--UPD--
Latest UI:
fine-tuning
model_view


corbt (commented):
We don't need to show the amount you saved with pruning -- just the total cost (with pruning rules applied) is fine.

Also, let's put the cost either directly above or directly below the "Start Training" button, which is where I'd expect to find it.

Kovbo:
Okay, I will experiment with this!
Displaying the amount near the "Start Training" button does not give you an instant price when changing the base model. A user needs to scroll the modal down each time to see the price.

It currently shows the amount with pruning only if there are applied pruning rules. In other cases, it shows the regular total price. Are there any other ways to assess the applied pruning rules?

Kovbo:
> We don't need to show the amount you saved with pruning -- just the total cost (with pruning rules applied) is fine.
> 
> Also, let's put the cost either directly above or directly below the "Start Training" button, which is where I'd expect to find it.

Okay, I will experiment with it!
Displaying the amount near the "Start Training" button does not give you an instant price when changing the base model. A user needs to scroll the modal down each time to see the price.

It currently shows the amount with pruning only if there are applied pruning rules. In other cases, it shows the regular total price. Are there any other ways to assess the applied pruning rules?

corbt:
I see what you mean -- yes, changing the model and immediately seeing the price change does seem nice. That said, the price of fine-tuning is usually much smaller than the price of inference, so I'm not sure the fine-tuning price is what the user cares most about when choosing a model anyway. Same thing with pruning rules -- the amount you save in costs while training with pruning rules is small compared to the amount you save with inference. So showing the savings there is not that important. (Does give me an idea though... we could include some kind of "inference calculator" where eg. we average the # of tokens in the test set and tell you how much you'd spend per 1000 calls on this model vs gpt-3.5 and gpt-4. Not a super high priority though!)

I'd still be in favor of putting the total down next to the button, because it's important that the user sees that before they click "Start Training". I'm ok if they have to scroll back and forth a bit if they want to see it while selecting models.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -93,16 +96,35 @@ export const datasetsRouter = createTRPCRouter({
...
       const totalTestingCount = await prisma.datasetEntry.count({
         where: {
-          datasetId: id,
+          datasetId,
           outdated: false,
           split: "TEST",
         },
       });
 
+      const { cost, inputCost, outputCost } = calculateCost(
```
arcticfly: One step that I forgot to mention is calculating the total number of epochs through which we'll run our training data. If the user has only a very small amount of data, we'll actually reuse it on multiple epochs.

If you search the codebase for `calculateNumEpochs`, you can see where we're already doing this in checkFineTuneStatus.task.ts.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -93,16 +96,35 @@ export const datasetsRouter = createTRPCRouter({
...
+      const { cost: costWithoutPruning } = calculateCost(
+        baseModel,
+        0,
+        trainingEntryStats.totalInputTokens,
+        trainingEntryStats.totalOutputTokens,
+      );
+
       return {
+        cost,
+        costWithoutPruning,
```
arcticfly: I think the only values we need to calculate and return from this function are those that change when you select/deselect a pruning rule or change a model. I'm open to feedback on this, but I think the other values we're returning from this function (`totalTestingCount` for instance) should be read from another query, as they currently are. That way we don't have to recalculate them every time we change the base model and need to recalculate cost.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,38 +55,39 @@ export const datasetsRouter = createTRPCRouter({
...
       const { projectId } = await prisma.dataset.findUniqueOrThrow({
-        where: { id },
+        where: { id: datasetId },
       });
 
       await requireCanViewProject(projectId, ctx);
 
-      const trainingEntryStats = await constructDatasetEntryFiltersQuery({ filters, datasetId: id })
+      // Ensure pruningRuleIds always contains at least one element, using a value that won't match anything
+      const safePruningRuleIds = pruningRuleIds.length > 0 ? pruningRuleIds : [uuidv4()];
```
arcticfly: Why can't this array be empty?

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,38 +55,39 @@ export const datasetsRouter = createTRPCRouter({
...
         .where("de.split", "=", "TRAIN")
         .where("de.output", "is not", null)
-        .select((eb) => [
-          sql<number>`count(*)::int`.as("numEntries"),
+        .leftJoin("PruningRuleMatch as prm", "prm.datasetEntryId", "de.id")
+        .leftJoin("PruningRule as pr", "prm.pruningRuleId", "pr.id")
+        .where("pr.id", "in", safePruningRuleIds)
+        .select([
+          sql<number>`count(distinct de.id)::int`.as("numEntries"),
           sql<number>`sum(de."inputTokens")::int`.as("totalInputTokens"),
```
arcticfly, line 81: I think we may run into a scenario in which the inputTokens of dataset entries with one or more matches are counted once for each match, instead of once for each dataset entry. The simplest solution would be to write one query to calculate the total tokens in all dataset entries, and another to calculate the total tokens saved by all pruning rule matches.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,38 +55,39 @@ export const datasetsRouter = createTRPCRouter({
...
+      const safePruningRuleIds = pruningRuleIds.length > 0 ? pruningRuleIds : [uuidv4()];
+
+      const trainingEntryStats = await constructDatasetEntryFiltersQuery({ filters, datasetId })
         .where("de.split", "=", "TRAIN")
         .where("de.output", "is not", null)
-        .select((eb) => [
-          sql<number>`count(*)::int`.as("numEntries"),
+        .leftJoin("PruningRuleMatch as prm", "prm.datasetEntryId", "de.id")
+        .leftJoin("PruningRule as pr", "prm.pruningRuleId", "pr.id")
+        .where("pr.id", "in", safePruningRuleIds)
```
arcticfly: You can actually make this where clause part of the join statement, and postgresql will be able to plan the query more efficiently.

https://kysely.dev/docs/examples/JOIN/complex-join

app/src/components/datasets/DatasetContentTabs/General/TrainingEntryMeter.tsx
```diff
@@ -1,12 +1,40 @@
...
+    {
+      datasetId: dataset?.id,
+      baseModel: splitProvider(props.selectedBaseModel),
+      filters,
+      pruningRuleIds: props.appliedPruningRuleIds,
+    },
+    { enabled: !!dataset },
+  );
+
+  console.log(stats.data);
```
arcticfly, line 34: remove

app/src/server/api/routers/datasets.router.ts
```diff
@@ -1,7 +1,6 @@
 import { z } from "zod";
 import { sql } from "kysely";
 import type { ComparisonModel } from "@prisma/client";
-
```
arcticfly: We actually like to leave these newlines in our import sections to separate imports from npm modules and local code. That can make it easier to find the import statement you're looking for.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
       await requireCanViewProject(projectId, ctx);
 
-      const trainingEntryStats = await constructDatasetEntryFiltersQuery({ filters, datasetId: id })
+      const datasetEntryStats = await constructDatasetEntryFiltersQuery({ filters, datasetId })
         .where("de.split", "=", "TRAIN")
         .where("de.output", "is not", null)
-        .select((eb) => [
-          sql<number>`count(*)::int`.as("numEntries"),
+        .select([
+          sql<number>`count(distinct de.id)::int`.as("numEntries"),
```
arcticfly: I don't believe this `distinct` operator is necessary.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
-            .whereRef("prm.datasetEntryId", "=", "de.id")
-            .where("prm.pruningRuleId", "in", selectedPruningRuleIds)
-            .leftJoin("PruningRule as pr", "prm.pruningRuleId", "pr.id")
-            .select(() => [sql<number>`sum(pr."tokensInText")::int`.as("totalMatchTokens")])
-            .as("totalMatchTokens"),
         ])
         .executeTakeFirst();
 
-      if (!trainingEntryStats) return error("No training data found");
+      if (!datasetEntryStats) return;
```
arcticfly, line 86: I think we should throw an error if we don't have any dataset entry stats.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
+
+      let totalMatchTokens = 0;
+
+      if (pruningRuleIds.length > 0) {
```
arcticfly, line 90: Can we remove this if statement?

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
+          .innerJoin("PruningRuleMatch", (join) =>
+            join
+              .onRef("PruningRuleMatch.datasetEntryId", "=", "de.id")
+              .on("PruningRuleMatch.pruningRuleId", "in", pruningRuleIds),
+          )
+          .leftJoin("PruningRule as pr", "PruningRuleMatch.pruningRuleId", "pr.id")
+          .select(sql<number>`sum(pr."tokensInText")::int`.as("totalMatchTokens"))
+          .executeTakeFirst();
+
+        totalMatchTokens = pruningRuleStats?.totalMatchTokens || 0;
```
arcticfly: Since the only thing we care about here is totalMatchTokens, I would set that directly using a `.then` statement.

i.e.
```
let totalMatchTokens = await
...
.executeTakeFirst()
.then((stats) => stats.totalMatchTokens);
```

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
+      );
+
+      const numEpochs = calculateNumEpochs(datasetEntryStats.numEntries);
 
       return {
-        matchingTrainingCount: trainingEntryStats.numEntries,
-        totalTestingCount,
+        cost: cost * numEpochs,
+        inputCost,
+        outputCost,
```
arcticfly: Do we use inputCost or outputCost for anything?

arcticfly (commented):
I personally like the design in which the cost is displayed above the buttons.

app/src/server/api/routers/datasets.router.ts
```diff
@@ -53,55 +55,67 @@ export const datasetsRouter = createTRPCRouter({
...
-      if (!trainingEntryStats) return error("No training data found");
+      if (!datasetEntryStats) return;
+
+      let totalMatchTokens = 0;
+
+      if (pruningRuleIds.length > 0) {
+        const pruningRuleStats = await constructDatasetEntryFiltersQuery({ filters, datasetId })
+          .where("de.split", "=", "TRAIN")
+          .where("de.output", "is not", null)
+          .innerJoin("PruningRuleMatch", (join) =>
```
arcticfly, line 92: I would put everything until this innerJoin into a variable named `baseQuery` and share it between both queries.

corbt (approved):
These changes look good to me!

PR #572, opened 2024-02-09
count fix for new openAi models

Kovbo:
A small tokens count fix for new OpenAi models.
If a model is not in a supported models list - we still try to get the inputTokens and outputTokens.

### DIFF
--- a/app/src/modelProviders/openai-ChatCompletion/frontend.ts
+++ b/app/src/modelProviders/openai-ChatCompletion/frontend.ts
@@ -2,6 +2,14 @@ const frontendModelProvider = {
   name: "OpenAI ChatCompletion",
 
   models: {
+    "gpt-4-0125-preview": {
+      name: "GPT-4 0125 Preview",
+      contextWindow: 128000,
+      promptTokenPrice: 0.00001,
+      completionTokenPrice: 0.00003,
+      speed: "fast",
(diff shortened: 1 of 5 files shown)


corbt (commented):
This is good. Can we also explicitly add the 0125 models as part of this PR so we get the correct costs for those?

corbt (approved):
Please add gpt-3.5-turbo-0125 as well, which is the new cheaper 3.5 variant.

Kovbo:
> This is good. Can we also explicitly add the 0125 models as part of this PR so we get the correct costs for those?

I added new models, but it appears that the package `gpt-tokens` does not support them. Wrote a temporary workaround.

There are places where we use old versions of turbo models for fine-tuning and evaluations. Should we update them to the new versions?

corbt:
Yep, anywhere we're using the old GPT-4-Turbo model let's update to the new one. I don't *think* we're using GPT-3.5-Turbo anywhere, but if we are we should hold off on updating that -- I've heard reports that the new GPT-3.5-Turbo actually is a weaker model.

Kovbo:
We support finetuning with gpt-3.5-turbo-1106.
As for the old GPT-4-Turbo, we use it as a comparison model for evaluations. The enum with this model is stored in the database (dataset.enabledComparisonModels), so I need advice on how to replace it.

PR #573, opened 2024-02-09
Document importing openpipe into a CommonJS project

arcticfly:
Enough projects still use CJS that we should document how to import OpenPipe.


corbt (approved):
Thanks!

PR #574, opened 2024-02-09
Record updated tag names on project

### LINKED ISSUE #570: Adding a new tag in an update request log request silently fails if column doesn't already exist
**Situation:** 
I want to use the OpenPipe client to add a new log tag but I noticed if that if the tag column doesn't already exist, the updateLogTag request silently fails. The response object says one request was matched (like it was successfully updated), but the tag wasn't actually added. 

In order to make it work, you need to ensure the tag column exists by adding it to a request from the OpenAI OpenPipe client. Once the column exists, then you can just use the updateLogTags normally. 

I think the fix should be to check if the tag doesn't exist when calling updateLogTags, then it creates the column and updates the tag.

arcticfly:
Fixes https://github.com/OpenPipe/OpenPipe/issues/570

We should also cache tag names that come in via the updateLogTags endpoint.

### DIFF
--- a/app/src/utils/recordRequest.ts
+++ b/app/src/utils/recordRequest.ts
@@ -236,7 +236,7 @@ async function createTags(projectId: string, loggedCallId: string, tags: Record<
 
   const tagNames = tagsToCreate.map((tag) => tag.name);
 
-  await recordTagNames(projectId, tagNames);
+  void recordTagNames(projectId, tagNames).catch((e) => captureException(e));
 }
 
 export async function recordTagNames(projectId: string, tagNames: string[]) {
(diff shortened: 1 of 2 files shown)


app/src/utils/recordRequest.ts
```diff
@@ -234,6 +234,12 @@ async function createTags(projectId: string, loggedCallId: string, tags: Record<
 
   await kysely.insertInto("LoggedCallTag").values(tagsToCreate).execute();
 
+  const tagNames = tagsToCreate.map((tag) => tag.name);
+
+  await recordTagNames(projectId, tagNames);
```
corbt: Can we just do `void recordTagNames(projectId, tagNames).catch(e => captureException(e)` to avoid the latency hit?

corbt (approved):

app/src/utils/recordRequest.ts
arcticfly: Good idea

PR #575, opened 2024-02-10
Input tokens are not output tokens

corbt:
This unfortunately led to some bad heuristics about sequence_len that caused some training jobs to fail. 😆


app/src/server/tasks/fineTuning/trainFineTune.task.ts
```diff
@@ -108,7 +109,7 @@ const trainModalFineTune = async (fineTuneId: string) => {
       };
       const prunedInputTokens = stringsToPrune?.length
         ? countLlamaInputTokens(input)
-        : dsEntry.outputTokens;
+        : dsEntry.inputTokens;
```
corbt, line 112: This is the only actual important change in this entire PR.

arcticfly (approved):

corbt:
Will run the following script in prod to fix the old/bad data:

```sql
update
  "FineTuneTrainingEntry" ftte
set
  "prunedInputTokens" = (
    select
      "inputTokens"
    from
      "DatasetEntry" de
    where
      de.id = ftte."datasetEntryId"
  )
where
  "prunedInputTokens" = "outputTokens"
  and ftte."fineTuneId" not in (
    select
      "fineTuneId"
    from
      "PruningRule"
    where
      "fineTuneId" is not null
  )
  ```

PR #576, opened 2024-02-10
readme update on how to use locally

latentloop07:
We got confused on how to use this locally. And spend lot of time debugging on how to use this exactly. So, we have added one section in readme on how to use this, when you are running this locally.


corbt (approved):
Thanks, super valuable contribution!

PR #580, opened 2024-02-12
Bugfix where training cost shows as $0.00

Kovbo:
The calculation issue arises because we recalculate tokens for each dataset entry as a background process in datasetEntries.createFromLoggedCalls function:
`countDatasetEntryTokens.enqueue()`

A simple solution is to check if any entries with `null` tokens exist.

If we consider adding more calculations to the dataset creation function, we could add a column to the dataset table that tracks if the dataset is processed.

### DIFF
--- a/app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
+++ b/app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
@@ -114,6 +114,9 @@ const FineTuneModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
     splitProvider(selectedBaseModel).baseModel === "mistralai/Mixtral-8x7B-Instruct-v0.1" &&
     isMissingBetaAccess;
 
+  const advancedConfigEnabled = splitProvider(selectedBaseModel).provider !== "openai";
+  const displayCostEnabled = splitProvider(selectedBaseModel).provider !== "openai";
+
   const numTrainingEntries = datasetEntries?.matchingTrainingCount || 0;
   const numTestingEntries = datasetEntries?.totalTestingCount || 0;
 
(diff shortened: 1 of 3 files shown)


arcticfly (commented):
Thanks for catching this. I think we should poll while calculations are in progress, right? You can find an example of how we do that in `EvaluationTable.tsx`.

Kovbo:
> Thanks for catching this. I think we should poll while calculations are in progress, right? You can find an example of how we do that in `EvaluationTable.tsx`.

Nice, I thought we did not have an easy way to implement it. Added.

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
```diff
@@ -105,6 +105,7 @@ const FineTuneModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
   const [trainingConfigOverrides, setTrainingConfigOverrides] = useState<
     Partial<AxolotlConfig> | undefined
   >();
+  const [calculationRefetchInterval, setCalculationRefetchInterval] = useState(0);
```
corbt, line 108: Think this could be replaced more elegantly with a `useMemo`.

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
```diff
@@ -327,78 +338,80 @@ const FineTuneModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
...
-                      </NumberInputStepper>
-                    </NumberInput>
-                  </VStack>
-                  <HStack>
-                    <Text fontWeight="bold">Number of Epochs</Text>{" "}
-                    <InfoCircle tooltipText="The total number of times the entire dataset is passed forward and backward through a neural network during training." />
-                  </HStack>
+                      <NumberInput
+                        step={0.0001}
+                        min={0.0001}
```
corbt: The minimum here should be 0 actually. It's possible someone will want to train with a learning rate several orders of magnitude smaller than our default.

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
arcticfly: Can it? The data we rely on to determine whether we should be polling or not is fetched from useDatasetTrainingCost, which is also the place where we need to specify our polling interval. I think a combination of useEffect and setState is the best way to get around this circular dependency, but I could be wrong.

Kovbo: I'm not sure I understand how to do it.
If I use `useMemo` to handle the state, it will depend on the query response: `price.data?.calculating`. The query response, at the same time, will depend on the interval from `useMemo` state, causing the "variable used before its declaration" error.
Maybe I misunderstood the comment?

corbt: Ah I see, yes that makes sense.

app/src/components/datasets/DatasetContentTabs/General/FineTuneButton.tsx
Kovbo: Done!

corbt (approved):

PR #581, opened 2024-02-12
Report to Sentry whenever a fine-tuning job fails

Kovbo:
Let me know if the function signature is not correct.


corbt (approved):
Thank you!

PR #587, opened 2024-02-13
Give our app access to the exports bucket

corbt:
We'll use this to upload the request logs export that the user asked for.

### DIFF
--- a/infra/src/app-user.ts
+++ b/infra/src/app-user.ts
@@ -1,0 +1,20 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import { nm } from "./helpers";
+import { writeExportBucket } from "./models";
+
+export const appUser = new aws.iam.User(nm("app"), {
+  name: nm("app"),
+});
+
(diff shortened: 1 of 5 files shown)


infra/src/app-user.ts
```diff
@@ -0,0 +1,20 @@
...
+import { nm } from "./helpers";
+import { writeExportBucket } from "./models";
+
+export const appUser = new aws.iam.User(nm("app"), {
+  name: nm("app"),
+});
+
+new aws.iam.UserPolicyAttachment(nm("app"), {
+  user: appUser.name,
+  policyArn: writeExportBucket.arn,
```
arcticfly, line 12: Do you plan to create a new bucket for exporting logs, or write them to the some one we use for exported models?

If the latter, I'd prefer a name change to reflect that the bucket no longer contains just models.

arcticfly (approved):
Good start

infra/src/app-user.ts
corbt: I think using the same bucket for both is fine. We can change the name at some point but didn't seem like a priority atm.

PR #588, opened 2024-02-14
Owner role added

Kovbo:
Added a migration that adds the `OWNER` ProjectUserRole.
Added a script to backfill project owners: `pnpm tsx src/server/scripts/backfillProjectOwners.ts`
Ownership cannot be transferred to another user. The owner cannot be removed from a project. The owner has the same permission as an admin and shares the same access control function `requireIsProjectAdmin`.

+Added permission for superadmins to view all projects.

### DIFF
--- a/app/src/components/projectSettings/ProjectUserTable.tsx
+++ b/app/src/components/projectSettings/ProjectUserTable.tsx
@@ -17,10 +17,6 @@ const ProjectUserTable = () => {
   const owner = selectedProject?.projectUsers.find((u) => u.role === "OWNER");
   const utils = api.useContext();
 
-  const canLeaveProject =
-    !isAdmin ||
-    !selectedProject ||
-    selectedProject.projectUsers.filter((u) => u.role === "ADMIN" || u.role === "OWNER").length > 1;
   const [projectUserToRemove, setProjectUserToRemove] = useState<ProjectUser | null>(null);
 
   const cancelInvitationMutation = api.users.cancelProjectInvitation.useMutation();
@@ -76,14 +72,10 @@ const ProjectUserTable = () => {
                   <Td fontSize={{ base: "xs", md: "sm" }}>{member.role}</Td>
                   <Td textAlign="end">
                     {(member.userId === session?.user?.id || isAdmin) &&
-                      member.userId !== owner?.userId &&
                       member.userId !== selectedProject.personalProjectUserId && (
                         <ConditionallyEnable
                           checks={[
-                            [
-                              member.userId !== session?.user?.id || canLeaveProject,
-                              "You are the last remaining admin. If you wish to leave, appoint another admin or delete the project.",
-                            ],
+                            [member.userId !== owner?.userId, "You are an owner of this project."],
                           ]}
                         >
                           <IconButton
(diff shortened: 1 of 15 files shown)


app/src/components/projectSettings/ProjectUserTable.tsx
```diff
@@ -14,13 +14,13 @@ const ProjectUserTable = () => {
...
   const isAdmin = useAccessCheck("requireIsProjectAdmin").access;
-
+  const owner = selectedProject?.projectUsers.find((u) => u.role === "OWNER");
   const utils = api.useContext();
 
   const canLeaveProject =
     !isAdmin ||
     !selectedProject ||
-    selectedProject.projectUsers.filter((u) => u.role === "ADMIN").length > 1;
+    selectedProject.projectUsers.filter((u) => u.role === "ADMIN" || u.role === "OWNER").length > 1;
```
arcticfly: Can we drop this condition and replace the `!isAdmin` condition with `u.role === "OWNER"`?

arcticfly (commented):
Looks good. Is there a migration script to convert the oldest admins to owners?

Kovbo:
> Looks good. Is there a migration script to convert the oldest admins to owners?

I added a script to do it:
pnpm tsx src/server/scripts/backfillProjectOwners.ts

As I remember, Postgres does not allow to add a new enum and use it in a single transaction, so we need to run the script manually.

app/src/components/projectSettings/ProjectUserTable.tsx
Kovbo: Hmm, we don't have a constraint on a database level that ensures that the project has an owner. It may happen that a project, for some reason, has only one admin. In this case, we should restrict leaving the project.
Or do you think that it is safe to remove this condition for the sake of simplicity?

arcticfly: The previous logic around Admins leaving a project is a little different from the new logic around Owners leaving a project. Since all projects will have a single owner from now on, users who are admins should be able to leave at any time, and owners should not be able to leave at all.

This check to ensure that another admin was on a project before allowing the previous one to leave made sense because a project could have multiple admins, and needed at least one. Since each project now has one and only one admin, this check no longer serves that purpose.

Kovbo: Yes, I understand the logic behind it.
There is a separate check that ensures that on owner cannot leave the project.
At the same time, there are no constraints on the database level that ensure that a project has an owner. We accidentally removed an owner by running some scripts. So this code ensures that at least one admin or owner should be associated with a project. 
Such cases are rare and noncritical. The question is whether we should ignore them to make the code look simpler.

arcticfly: I think we should avoid deleting project owners, and assume that every project has an owner. If a project ends up without an owner through an error in our logic, billing will also be in a broken state.

We can and should avoid this edge case.

Kovbo: I changed the code. It looks simpler if we do not think about such edge cases.

app/src/components/projectSettings/ProjectUserTable.tsx
```diff
@@ -79,10 +75,7 @@ const ProjectUserTable = () => {
                       member.userId !== selectedProject.personalProjectUserId && (
                         <ConditionallyEnable
                           checks={[
-                            [
-                              member.userId !== session?.user?.id || canLeaveProject,
-                              "You are the last remaining admin. If you wish to leave, appoint another admin or delete the project.",
-                            ],
+                            [member.userId !== owner?.userId, "You are an owner of this project."],
```
arcticfly, line 78: Nit: You are the owner of this project.

arcticfly (approved):
Looks good!

PR #589, opened 2024-02-15
Merge on CPU

arcticfly:
User's aren't time-sensitive on model merging and export right now, and doing it on a CPU gets around memory constraints while saving money.


corbt (approved):
You just snuck in that bump to h100s didn't ya 😆

PR #592, opened 2024-02-15
Evaluate view changed

Kovbo:
Task:
Current behavior:
In the evaluation table, we make each row as tall as the tallest output. To condense long inputs into this shorter space, we show the beginning of the input log, along with a button to expand it.

Changes:
Instead of showing the beginning of long inputs by default, we should show the end of them. The last message in an input is more likely to be different and relevant to the final output than the first one, which is often a system message.

We should continue collapsing and expanding inputs, but we should cap the max height to some reasonable amount, perhaps max(600px, height_of_tallest_output). If an input is taller than our max height, it should be scrollable.

Inputs that are not expanded should not be scrollable.

Previous UI:
prev

Current UI:


corbt:
Let's remove the "ID: " line entirely, haven't had to use it for much in practice and uses a lot of space with these changes. We can bring back permalinks to rows sometime later.

corbt (approved):

Kovbo:
> Let's remove the "ID: " line entirely, haven't had to use it for much in practice and uses a lot of space with these changes. We can bring back permalinks to rows sometime later.

Changed:

PR #595, opened 2024-02-16
Disable the fine-tune button correctly

corbt:
Because we were setting the `isDisabled` prop explicitly, the checks in `ConditionallyEnable` were being bypassed, and users could attempt (and fail) to create fine-tunes in bad states like eg. trying to fine-tune GPT-3.5 without an API key.


Kovbo (approved):
Nice, I thought `ConditionallyEnable` hides the button based on a condition, not disables it. Has to check the code :)

PR #598, opened 2024-02-17
No GPU autoscaling

corbt:
For some reason anyscale really doesn't want to autoscale *down* and autoscaling up is so slow it's useless anyway, so let's disable autoscaling for the moment.


inference/pyproject.toml
```diff
@@ -1,14 +1,12 @@
...
 version = "0.1.0"
 description = ""
 authors = ["Kyle Corbitt <kyle@corbt.com>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.9"
-anyscale = "^0.5.171"
-ray = {extras = ["serve"], version = "^2.9.0"}
+python = "^3.12"
```
arcticfly: Why are we specifying version 3.9 in .tool-versions and 3.12 in pyproject.toml?

arcticfly (approved):

inference/pyproject.toml
corbt: Good catch; needs to be 3.9 because Ray doesn't have a 3.12 dist available. Updated to make this consistent.

PR #599, opened 2024-02-17
Anyscale and Fireworks streaming added

Kovbo:
I think I figured out how to add support of streams to Anyscale and Fireworks :)  I still have a limited understanding of streaming in general, so there might be some mistakes.
Aniscale has a very similar API response compared to OpenAI. The only difference is the `object` property and the `usage` property. Fireworks has a similar yet different response structure.

As for Modal, do we still support it? How do I train a model to test it?

Also, I do not check if it is a function call. Can we check the input to see if it is a function call with streaming and throw an error, or we need to parse the output?

This pull request also has a migration that updates gpt-4 turbo as a comparison model.


client-libs/typescript/src/openai/mergeChunks.ts
```diff
@@ -13,7 +14,7 @@ const omit = <T extends Record<string, unknown>, K extends keyof T>(
 
 export default function mergeChunks(
   base: ChatCompletion | null,
-  chunk: ChatCompletionChunk,
+  chunk: ChatCompletionChunk & { usage?: CompletionUsage },
```
Kovbo, line 17: Both Anyscale and Fireworks have a usage property that is not compatible with OpenAI chunks. Should we introduce a custom type for chunks?

client-libs/typescript/src/openai/http.test.ts
```diff
@@ -11,16 +12,76 @@ test("fetches non-streamed output", async () => {
...
+          expect(choice).toHaveProperty("logprobs");
+          expect(choice).toHaveProperty("finish_reason");
+        });
+      });
+    }
+  }
+
+  await processStream();
+  expect(chunkCount).toBeGreaterThan(1);
+}, 200000);
```
Kovbo, line 87: Anyscale cold start takes a really long time.

app/src/utils/comparisonModels.ts
```diff
@@ -14,8 +14,9 @@ export const isComparisonModelName = (modelName: string) =>
 
 export const COMPARISON_MODEL_NAMES: Record<ComparisonModel, string> = {
   GPT_3_5_TURBO: "gpt-3.5-turbo-1106",
-  GPT_4_1106_PREVIEW: "gpt-4-1106-preview",
+  GPT_4_0125_PREVIEW: "gpt-4-0125-preview",
```
corbt: As long as we're here let's rename these to be more human-friendly

"GPT-3.5 Turbo (-1106)"
"GPT-4 Turbo (-0125)"
"GPT-4 (0613)"
"GPT-4 Turbo (-1106)"

app/src/modelProviders/fine-tuned/getAnyscaleCompletion.ts
```diff
@@ -79,3 +99,26 @@ export async function getAnyscaleCompletion(
...
+
+  return new Stream(iterator, controller);
+}
+
+function transformChunk(chunk: any) {
+  return {
+    id: chunk.id,
+    object: "chat.completion.chunk",
+    created: chunk.created,
+    model: chunk.model,
```
corbt: In both this case and the Fireworks case we should explicitly set the model as the model the user requested (ie. `openpipe:some-slug`) instead of passing through the anyscale/fireworks model name, as those are different and only implementation details.

client-libs/typescript/src/openai/http.test.ts
```diff
@@ -11,16 +12,76 @@ test("fetches non-streamed output", async () => {
...
     }),
   });
 
   const data = await response.json();
+  console.log(data);
   expect(data.choices[0].message.content).toBeDefined();
-});
+}, 200000);
+
+test.only("fetches streamed output", async () => {
```
corbt: Nit: remove the `.only` before merging

client-libs/typescript/src/openai/http.test.ts
```diff
@@ -11,16 +12,76 @@ test("fetches non-streamed output", async () => {
...
+        expect(chunkJson.object).toEqual("chat.completion.chunk");
+        expect(chunkJson).toHaveProperty("created");
+        expect(chunkJson).toHaveProperty("model");
+        expect(chunkJson).toHaveProperty("choices");
+        expect(Array.isArray(chunkJson.choices)).toBeTruthy();
+        chunkJson.choices.forEach((choice: Record<string, any>[]) => {
+          expect(choice).toHaveProperty("index");
+          expect(choice).toHaveProperty("delta");
+          expect(choice).toHaveProperty("logprobs");
+          expect(choice).toHaveProperty("finish_reason");
```
corbt: we should make this more explicit that the `finish_reason` is one of the reasons supported by openai

client-libs/typescript/src/openai/http.test.ts
corbt: Let's add another test that runs similar checks but uses the official OpenAI SDK to parse the streams (https://www.npmjs.com/package/openai#streaming-responses).

Also I'm assuming the answer is "yes", but I'm assuming you've visually inspected the chunks returned by all of openai,anyscale,fireworks to make sure eg. the `finish_reason` is of the types openai supports.

client-libs/typescript/src/openai/http.test.ts
corbt: weird -- it actually should have a GPU running in the dev env so shouldn't be *that* long.

client-libs/typescript/src/openai/mergeChunks.ts
corbt: I think returning usage is fine but I probably wouldn't document that we do, since we might add more providers in the future that stick closer to the OpenAI spec and don't include it.

corbt (commented):
Very excited for this!

For detecting whether the input is a tool call, we can check whether it has the `tools` or `functions` key set. If it has entries in either of those arrays then there's the possibility that the model will call a tool, so we should throw an error for now if those have entries and `streaming` is also set to `true`.

app/src/utils/comparisonModels.ts
Kovbo: We rely on values in some places because it is compatible with openAi `model` field. I renamed it by adding an object as a value, that has an OpenAI compatible `name` and a human-readable `slug`.

app/src/modelProviders/fine-tuned/getAnyscaleCompletion.ts
Kovbo: Changed it!

client-libs/typescript/src/openai/http.test.ts
Kovbo: Added this explicit check.

client-libs/typescript/src/openai/http.test.ts
Kovbo: Added a direct call, it passes the test.

app/src/utils/comparisonModels.ts
```diff
@@ -3,29 +3,36 @@ import { ComparisonModel } from "@prisma/client";
...
-  Object.values(COMPARISON_MODEL_NAMES).includes(modelName);
+  Object.values(COMPARISON_MODEL_NAMES).some((model) => model.name === modelName);
 
-export const COMPARISON_MODEL_NAMES: Record<ComparisonModel, string> = {
-  GPT_3_5_TURBO: "gpt-3.5-turbo-1106",
-  GPT_4_1106_PREVIEW: "gpt-4-1106-preview",
-  GPT_4_0613: "gpt-4-0613",
+export const COMPARISON_MODEL_NAMES: Record<ComparisonModel, { name: string; slug: string }> = {
+  GPT_3_5_TURBO: { name: "gpt-3.5-turbo-1106", slug: "GPT-3.5 Turbo (-1106)" },
+  GPT_4_0613: { name: "gpt-4-0613", slug: "GPT-4 (0613)" },
```
corbt: The slug isn't consistent here. Should be (-0613).

app/src/modelProviders/fine-tuned/getFireworksCompletion.ts
```diff
@@ -36,6 +36,12 @@ export async function getFireworksCompletion(
     throw new Error("FIREWORKS_API_KEY is required for Fireworks completions");
   }
 
+  if ((input.functions || input.tools) && input.stream) {
+    throw new Error(
+      `Error: Streaming completions are not supported for function calls. Model: ${fineTune.slug}`,
```
corbt: "We don't currently support streaming for function calls. Please open an issue if you need this functionality! https://github.com/openpipe/openpipe"

corbt:
Couple small tweaks then this should be ready to ship!

Kovbo:
> Couple small tweaks then this should be ready to ship!

All done!

corbt (approved):
People are going to love it!

PR #600, opened 2024-02-17
Add nodes

corbt:


app/prisma/schema.prisma
```diff
@@ -722,3 +857,204 @@ model ExportWeightsRequest {
...
+
+    @@unique([parentNodeEntryId, dataChannelId])
+    @@index([nodeId, persistentId])
+    @@index([nodeId, inputHash, persistentId])
+    @@index([loggedCallId, dataChannelId])
+    @@index([parentNodeEntryId, dataChannelId])
+    @@index([nodeId, status, updatedAt])
+}
+
+model DatasetEntryInput {
```
corbt, line 1000: Maybe rename to `LLMInput` and `LLMOutput`?

app/src/server/tasks/nodes/processNodes/processNode.task.ts
```diff
@@ -0,0 +1,263 @@
...
+export const processNode = defineTask<ProcessNodeJob>({
+  id: "processNode",
+  handler: async (task) => {
+    console.log(task);
+
+    const { nodeId, nodeType, invalidateData } = task;
+    if (invalidateData) {
+      await invalidateNodeEntries(nodeId);
+    }
+    if (
```
corbt: remove condition?

app/src/server/tasks/nodes/processNodes/processNode.task.ts
```diff
@@ -0,0 +1,269 @@
...
+                if (nodeProperties.cacheMatchFields?.includes("incomingDEIHash")) {
+                  ands.push(eb("ne.inputHash", "=", `cpe.incomingDEIHash`));
+                }
+                if (nodeProperties.cacheMatchFields?.includes("incomingDEOHash")) {
+                  ands.push(eb("ne.outputHash", "=", `cpe.incomingDEOHash`));
+                }
+
+                return eb.and(ands);
+              }),
+            ),
```
corbt, line 187: Don't make the callee return all the same stuff you gave it

app/src/components/datasets/DatasetContentTabs/General/DownloadButton.tsx
```diff
@@ -1,181 +0,0 @@
...
-        isDisabled={selectedIds.size === 0}
-      />
-      <ExportDatasetEntriesModal disclosure={disclosure} />
-    </>
-  );
-};
-
-export default ExportButton;
-
-const ExportDatasetEntriesModal = ({ disclosure }: { disclosure: UseDisclosureReturn }) => {
```
arcticfly, line 55: @corbt What do you think about exporting datasets?

app/src/server/api/external/v1Api/procedures/unstableDatasetEntryCreate.procedure.ts
```diff
@@ -24,7 +26,7 @@ export const unstableDatasetEntryCreate = openApiProtectedProc
   })
   .input(
     z.object({
-      datasetId: z.string(),
+      archiveId: z.string(),
```
arcticfly, line 29: We should update our docs on this change. Is anyone using this route right now?

app/prisma/schema.prisma
```diff
@@ -744,3 +894,198 @@ model OngoingRequest {
...
+    @@index([nodeId, persistentId])
+    @@index([nodeId, inputHash, persistentId])
+    @@index([loggedCallId, dataChannelId])
+    @@index([parentNodeEntryId, dataChannelId])
+    @@index([nodeId, status, updatedAt])
+    @@index([persistentId, nodeId])
+}
+
+model DatasetEntryInput {
+    tool_choice     Json?
```
corbt, line 1001: I have a weak preference for combining these into a single input JSON field to give us schema/version flexibility.

app/src/server/utils/nodes/hashNode.ts
```diff
@@ -0,0 +1,124 @@
...
+      messages: JSON.stringify(messages),
+      response_format: response_format ? JSON.stringify(response_format) : undefined,
+    })
+    .onConflict((oc) => oc.columns(["hash"]).doNothing())
+    .execute();
+
+  return inputHash;
+};
+
+export const hashDatasetEntryOutput = ({
```
corbt, line 85: These functions should either (a) be more strongly typed to only accept a ChatCompletionMessage, or (b) zod-validate their input before saving.

If we don't do that and end up accidentally saving something in here that's the wrong shape, we'll start getting errors all over the place when we try to read it back out and access fields that don't exist.

Same thing for the input as well.

corbt: Ah I think you can actually use the zod validators you've already defined for `typedDatasetEntryInput` and `typedDatasetEntryOutput`?

app/src/server/utils/nodes/nodeProperties/monitorProperties.ts
```diff
@@ -0,0 +1,131 @@
...
+      }),
+      prisma.nodeEntry.createMany({
+        data: nodeEntriesToCreate,
+        skipDuplicates: true,
+      }),
+    ]);
+  },
+};
+
+const calculateSampleRateHash = (sampleRate: number) => {
```
corbt, line 112: So this is how I was expecting to implement this but I am still super low-confidence it will work lol. But we can test later I guess, none of this goes live immediately?

app/prisma/schema.prisma
arcticfly: I'd prefer to keep the names as they are. Since they'll sometimes be generated through manual editing, including LLM in the name could be misleading.

app/prisma/schema.prisma
arcticfly: I think that's a decent idea, but I'd prefer to do it outside of this migration. That will make it easier to find the cause of any new bugs that might appear.

app/src/server/api/external/v1Api/procedures/unstableDatasetEntryCreate.procedure.ts
corbt: No one is using it in prod, I think if we regenerate the docs that's good enough.

app/src/components/datasets/DatasetContentTabs/General/DownloadButton.tsx
corbt: Yes, we should enable it.

app/src/server/utils/nodes/hashNode.ts
arcticfly: I think improving the types of these functions' arguments and using a zod validator elsewhere in the code before calling these functions makes the most sense. I originally chose the current argument types in order to minimize type casting, but I agree that stronger typing is the better option.

app/src/server/utils/nodes/nodeProperties/monitorProperties.ts
arcticfly: Correct

arcticfly (approved):

PR #610, opened 2024-02-20
Invoices copy imrovement

Kovbo:
Change list:

- Add a “Pay Now” button at the bottom of each invoice in the “pending” state.
    -  When a user clicks this button, or clicks the “Pay” button from the invoices list, we should open the modal and let them enter their card details immediately. Currently there’s an error that pops up if you don’t have a payment method on file.
-  Rename the “Pending” state to “Unpaid”
-  Email copy changes
    -  Title should be “OpenPipe Invoice Dec 2023 - Jan 2024”, not “OpenPipe Usage Dec 2023 - Jan 2024”
    -  If the invoice is just for one month (ie they created their account in January) the title should end in “OpenPipe Invoice Jan 2024” instead of “OpenPipe Invoice Jan 2024 - Jan 2024”
    -  If the invoice requires payment, the button at the bottom should be “Pay Now”, not “Details”
    -  “Billing Total” should always include dollars and cents, ie. “$0.00”, not just “$0”
    

I created a script that renames the billing period if a project was created in January, and currently have the billing period as "Jan 2024 - Jan 2024":
`pnpm tsx src/server/scripts/renameBillingPeriods.ts`


corbt (approved):

corbt:
Thank you @Kovbo !

PR #612, opened 2024-02-20
Confirm dataset deletion, email finetune success, UI fixes.

Kovbo:
Task: Require users to confirm by typing the dataset name before deleting a dataset.

UI:
datasetDelete

Task: Send success emails for fine tune creation.

UI:
email

Task: Avoid models table overflow in usage view.
UI + responsive improvements:
usage-mobile

Task: Usage numbers should always include two decimals
Done.

### DIFF
--- a/app/src/server/emails/templates/layout/Highlight.tsx
+++ b/app/src/server/emails/templates/layout/Highlight.tsx
@@ -5,7 +5,7 @@ interface Props {
   style?: CSSProperties;
   children: ReactNode;
 }
-export default function Headlight({ children, style }: Props) {
+export default function Headlight({ style, children }: Props) {
   return <Section style={{ ...style, ...dark }}>{children}</Section>;
 }
 
--- a/app/src/server/emails/templates/layout/Highlight.tsx
+++ b/app/src/server/emails/templates/layout/Highlight.tsx
@@ -1,0 +1,16 @@
+import { Section } from "@react-email/components";
+import React, { CSSProperties, ReactNode } from "react";
+
+interface Props {
+  style?: CSSProperties;
+  children: ReactNode;
+}
+export default function Headlight({ children, style }: Props) {
+  return <Section style={{ ...style, ...dark }}>{children}</Section>;
+}
+
+const dark = {
+  padding: "24px",
+  backgroundColor: "#f2f3f3",
+  borderRadius: "4px",
+};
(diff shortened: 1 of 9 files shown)


app/src/server/tasks/fineTuning/checkFineTuneStatus.task.ts
```diff
@@ -83,6 +85,20 @@ export const checkFineTuneStatus = defineTask({
               },
             });
 
+            // Notify the owner that the model has been trained
+            const project = await prisma.project.findUniqueOrThrow({
+              where: { id: typedFT.projectId },
+            });
+            await sendToOwner(typedFT.projectId, (email: string) =>
```
Kovbo, line 92: Do we notify a project owner or all admins?

corbt: Actually let's track the user who created the fine-tune and alert just that user in this case. Shouldn't be that hard to add and will be a better ux.

Kovbo: I can do it by adding a column to the fineTunes table in order to track a user.
Do we need to backfill the existing models with users?

corbt: Nah let's just add the column going forward, no need to backfill.

app/src/server/emails/templates/layout/Highlight.tsx
```diff
@@ -0,0 +1,16 @@
+import { Section } from "@react-email/components";
+import React, { CSSProperties, ReactNode } from "react";
+
+interface Props {
+  style?: CSSProperties;
+  children: ReactNode;
+}
+export default function Headlight({ style, children }: Props) {
```
arcticfly: Nit: Headlight -> Highlight

app/src/server/emails/templates/layout/InvoiceDescription.tsx
```diff
@@ -1,6 +1,7 @@
-import { Column, Row, Section } from "@react-email/components";
+import { Column, Row } from "@react-email/components";
 import React, { CSSProperties } from "react";
 import { Text } from ".";
+import Headlight from "./Highlight";
```
arcticfly: Headlight -> Highlight

arcticfly (approved):

PR #605, opened 2024-02-18
[Snyk] Upgrade lucide-react from 0.265.0 to 0.316.0

corbt:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade lucide-react from 0.265.0 to 0.316.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **52 versions** ahead of your current version.
- The recommended version was released **23 days ago**, on 2024-01-26.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>lucide-react</b></summary>
    <ul>
      <li>
        <b>0.316.0</b> - 2024-01-26
      </li>
      <li>
        <b>0.315.0</b> - 2024-01-25
      </li>
      <li>
        <b>0.314.0</b> - 2024-01-21
      </li>
      <li>
        <b>0.313.0</b> - 2024-01-21
      </li>
      <li>
        <b>0.312.0</b> - 2024-01-17
      </li>
      <li>
        <b>0.311.0</b> - 2024-01-16
      </li>
      <li>
        <b>0.310.0</b> - 2024-01-16
      </li>
      <li>
        <b>0.309.0</b> - 2024-01-10
      </li>
      <li>
        <b>0.308.0</b> - 2024-01-09
      </li>
      <li>
        <b>0.307.0</b> - 2024-01-06
      </li>
      <li>
        <b>0.306.0</b> - 2024-01-05
      </li>
      <li>
        <b>0.305.0</b> - 2024-01-05
      </li>
      <li>
        <b>0.304.0</b> - 2024-01-04
      </li>
      <li>
        <b>0.303.0</b> - 2023-12-28
      </li>
      <li>
        <b>0.302.0</b> - 2023-12-25
      </li>
      <li>
        <b>0.301.0</b> - 2023-12-24
      </li>
      <li>
        <b>0.300.0</b> - 2023-12-22
      </li>
      <li>
        <b>0.299.0</b> - 2023-12-20
      </li>
      <li>
        <b>0.298.0</b> - 2023-12-16
      </li>
      <li>
        <b>0.297.0</b> - 2023-12-15
      </li>
      <li>
        <b>0.296.0</b> - 2023-12-15
      </li>
      <li>
        <b>0.295.0</b> - 2023-12-13
      </li>
[...]

PR #602, opened 2024-02-17
[Snyk] Upgrade node-fetch from 2.6.12 to 2.7.0

corbt:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade node-fetch from 2.6.12 to 2.7.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **2 versions** ahead of your current version.
- The recommended version was released **6 months ago**, on 2023-08-23.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>node-fetch</b></summary>
    <ul>
      <li>
        <b>2.7.0</b> - <a href="https://snyk.io/redirect/github/node-fetch/node-fetch/releases/tag/v2.7.0">2023-08-23</a></br><h1><a href="https://snyk.io/redirect/github/bitinn/node-fetch/compare/v2.6.13...v2.7.0">2.7.0</a> (2023-08-23)</h1>
<h3>Features</h3>
<ul>
<li><code>AbortError</code> (<a href="https://snyk.io/redirect/github/bitinn/node-fetch/issues/1744" data-hovercard-type="pull_request" data-hovercard-url="/node-fetch/node-fetch/pull/1744/hovercard">#1744</a>) (<a href="https://snyk.io/redirect/github/bitinn/node-fetch/commit/9b9d45881e5ca68757077726b3c0ecf8fdca1f29">9b9d458</a>)</li>
</ul>
      </li>
      <li>
        <b>2.6.13</b> - <a href="https://snyk.io/redirect/github/node-fetch/node-fetch/releases/tag/v2.6.13">2023-08-18</a></br><h2><a href="https://snyk.io/redirect/github/bitinn/node-fetch/compare/v2.6.12...v2.6.13">2.6.13</a> (2023-08-18)</h2>
<h3>Bug Fixes</h3>
<ul>
[...]

PR #613, opened 2024-02-20
Dependencies update

Kovbo:
Dependencies list:
https://www.notion.so/opipe/f97da538f2db46f5a5bcfdff9452f2c2?v=66f18e7dd47e4a9b8efda2f411d7b3c7&p=ab07b5b4a7314ff695a94340269e6863&pm=s

Removed @sendinblue/client. We no longer use this package?

Python dependencies are still pending.
A report says "pip-fastapi <= 0.109.0". But it seems we already use a higher version.

Addressed these PR:
https://github.com/OpenPipe/OpenPipe/pull/602
https://github.com/OpenPipe/OpenPipe/pull/605


corbt (approved):
Thank you!

PR #615, opened 2024-02-21
Fireworks rate limiting

corbt:
Ensure we never send more than 30 requests to Fireworks as part of evals at once, and that we send requests at a maximum rate of 10 rps.

Also adds a general `rateLimit` abstraction that we may be able to reuse for similar types of soft limits in the future. It uses the token bucket strategy -- leaky bucket would probably be a better fit semantically, but would require some pretty major changes to the way we queue/process background jobs so decided to just go with the simple approach instead.

### DIFF
--- a/app/src/server/tasks/generateTestSetEntry.task.ts
+++ b/app/src/server/tasks/generateTestSetEntry.task.ts
@@ -212,8 +212,7 @@ export const generateTestSetEntry = defineTask<GenerateTestSetEntryJob>({
           await prisma.fineTuneTestingEntry.update({
             where: { modelId_datasetEntryId: { modelId, datasetEntryId } },
             data: {
-              errorMessage:
-                "Pausing completion processing to avoid rate limiting, will retry soon.",
+              errorMessage: "Pending",
             },
           });
         } else {
--- a/app/src/server/tasks/generateTestSetEntry.task.ts
+++ b/app/src/server/tasks/generateTestSetEntry.task.ts
@@ -20,6 +20,8 @@ import {
 import { getOpenaiCompletion } from "../utils/openai";
 import defineTask from "./defineTask";
 import { queueHeadToHeadEvalJobsForTestingEntry } from "./evaluateTestSetEntries.task";
+import { RateLimitError } from "openai";
+import { fireworksTestSetLimit } from "~/utils/rateLimits";
 
 export type GenerateTestSetEntryJob = {
   modelId: string;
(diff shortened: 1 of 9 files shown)


app/src/server/tasks/generateTestSetEntry.task.ts
```diff
@@ -20,6 +20,8 @@ import {
 import { getOpenaiCompletion } from "../utils/openai";
 import defineTask from "./defineTask";
 import { queueHeadToHeadEvalJobsForTestingEntry } from "./evaluateTestSetEntries.task";
+import { RateLimitError } from "openai";
```
arcticfly, line 23: Nit: import order

app/src/utils/rateLimits.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    await rateLimit(key, rateLimitConfig);
+
+    // Wait for 1 second to allow the bucket to refill
+    await sleep(1000);
+    const result = await rateLimit(key, rateLimitConfig);
+    expect(result).toBe(true);
+  });
+
+  it.only("calls the fireworksTestSetLimit function", async () => {
```
arcticfly: Remove it.only

app/src/utils/rateLimits.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const currentTokenCount = sql`least(${config.capacity}, "RateLimit".tokens + ${tokensToAdd})`;
+
+  const query = sql<{ allowed: boolean; tokensLeft: number }>`
+    insert into "RateLimit"(key, tokens, allowed, "lastUpdated")
+    values (${key}, ${config.capacity} - 1, true, now())
+    on conflict (key) do update set
+      tokens = greatest(0, ${currentTokenCount} - 1),
+      allowed = round(${currentTokenCount}) > 0,
+      "lastUpdated" = now()
```
arcticfly: By this logic, if we get hammered with requests, won't we keep updating "lastUpdated" while not actually sending any requests at all?

We should avoid setting "lastUpdated" if currentTokenCount is 0.

app/src/utils/rateLimits.ts
```diff
@@ -0,0 +1,32 @@
...
+    returning allowed, tokens as "tokensLeft";
+  `;
+  const result = await query.execute(kysely);
+  console.log(result.rows[0]);
```
arcticfly: Remove console

arcticfly (changes requested):

app/src/server/tasks/generateTestSetEntry.task.ts
corbt: We can fix this by adding an eslint rule to automatically sort imports after you get your changes merged.

arcticfly (approved):
Optimistically approving

PR #619, opened 2024-02-21
Stripe webhook fix

Kovbo:
Small fix addressing the issue when a subscription payment does not have an associated invoice, but still creates a webhook.


corbt (approved):
Thanks!

PR #622, opened 2024-02-22
Bump python openai sdk version

### LINKED ISSUE #609: Strict version constraint `openai = "~1.7.0"` in Python client prevents updating packages
I've been trying out OpenPipe (love the idea!), but today I got into a situation where I needed to update the `openai` package (installing `langchain-openai` requires me to do so).

I can see how it might be a lot of work for you to keep up-to-date with the latest `openai` version, but since my need for using the latest `openai` package is greater than my need to try out OpenPipe, I might have to drop OpenPipe. 

Before I do so, I wanted to reach out and check what the status is on that version constraint, and if it can be updated/relaxed.

arcticfly:
Update python SDK version to fix #609

### DIFF
--- a/client-libs/python/pyproject.toml
+++ b/client-libs/python/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "openpipe"
-version = "4.4.1"
+version = "4.4.2"
 description = "Python client library for the OpenPipe service"
 authors = ["Kyle Corbitt <kyle@openpipe.ai>"]
 license = "Apache-2.0"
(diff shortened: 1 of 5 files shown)


client-libs/python/pyproject.toml
```diff
@@ -13,7 +13,7 @@ python = "^3.8"
 httpx = "^0.24.1"
 attrs = "^23.1.0"
 python-dateutil = "^2.8.2"
-openai = "~1.7.0"
+openai = ">= 1.7, <= 1.12"
```
ErikBjare: You probably mean: 
```suggestion
openai = ">= 1.7, < 1.13"
```
Otherwise, you'd be stuck with `1.12.0` (and not get patch versions like `1.12.1`).

arcticfly: There are definitely tradeoffs here. In the past OpenAI has released patch versions that actually broke our ingestion workflow (breaking changes aren't guaranteed to be minor or major version updates). But since OpenAI's SDK has been reasonably stable for a couple months, I'm willing to take a chance and include `1.12` patch versions!

Let me update the dependency requirements.

Kovbo (approved):

PR #625, opened 2024-02-22
[WIP] Setup new Auth0 provider for easy local development login

rajshirolkar:
This PR sets up a simple Auth0 provider with a free dev instance of an Auth0 app. Users can locally login with Auth0 by entering the credentials

email : dev@example.com
password : Test#1234

Alternatively a user can also Sign Up as a new user and that flow would also work out of the box.

Note : I still have to update the Readme and the `.env` file with the credentials for Auth0 when we create a new Auth0 application for this.

### DIFF
--- a/app/src/components/nav/AppShell.tsx
+++ b/app/src/components/nav/AppShell.tsx
@@ -114,7 +114,11 @@ const NavSidebar = () => {
               as={ChakraLink}
               justifyContent="start"
               onClick={() => {
-                signIn("github").catch(console.error);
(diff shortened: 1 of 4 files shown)


corbt:
Thanks for the code. To be honest my spidey sense is still tingling with this one... I feel like if I were running this repo and had to connect to a 3rd party oauth provider I'd feel like something was wrong and wonder what other information I'm reporting to third-parties just by running this locally. I think we need a fully-local solution here for this to really meet the needs of our users.

PR #626, opened 2024-02-23
Admin UI

Kovbo:
Adding admin dashboard that allows searching projects by name and slug, by user name and email, by model slug and baseMode.
Admin dashboard

Do we need separate tables for Users and Models?


.gitignore
```diff
@@ -4,4 +4,6 @@
 node_modules/
 *.tsbuildinfo
 dist/
-tmp/
\ No newline at end of file
+tmp/
+.DS_Store
+client-libs/.DS_Store
```
Kovbo, line 9: Git keeps adding this file in these two folders.

corbt (approved):
I think starting with one table as you have here is good enough for the moment. I'm excited to have this available!

PR #629, opened 2024-02-23
Rate limit

Kovbo:
Plan: use concurrency-based rate limiting, because that forms a natural kind of back-pressure (under higher load we will take longer to respond). Also makes it easier for us to have one limit that applies to models of different sizes since larger models will naturally respond slower.

Task:
- Add a  `rateLimit` field to all projects, with a default value of 5.
    - Existing projects should have this set to 1000 to start as part of the migration — we can lower it later
- Create a new table `OngoingRequest`. It should only have the fields `id`, `projectId`, `createdAt`. Index on [`projectId`, `createdAt` ]
- When a new inference request comes in for a project, we should check if there are fewer than `rateLimit` requests in the `OngoingRequest` table for that project that were started in the last 10 minutes (requests older than 10 minutes we can assume are hung jobs and can be disregarded).
    - If there are > `rateLimit` requests already in flight, the SQL query should return `false` and we should throw a 429 error.
    - If there are < `rateLimit` requests already in flight, the SQL query should insert a new request in the `OngoingRequest` table and return its id
    - At the end of generation (whether successful or not) we should always delete the request from the `OutgoingRequest` table.
- Add to the project “Settings” page a (read-only) section showing the project rate limit as well as an explanation of how rate-limiting works.

UI:

____

Added admin dashboard UI to change rate limits.

Also, this pull request contains logic to notify the fine-tuned creator about completed training.


app/src/pages/p/[projectSlug]/settings/index.tsx
```diff
@@ -180,6 +181,19 @@ export default function Settings() {
...
+            <VStack alignItems="flex-start">
+              <Subtitle>Rate Limit</Subtitle>
+              <HStack>
+                <Text color="gray.500">Concurrent requests:</Text>
+                <Text fontWeight="bold">{selectedProject?.rateLimit}</Text>
+                <InfoCircle tooltipText="The limits are based by the number of concurrent requests. There is no limit on the total number of requests you can send, but rather on how many you can send at the same time." />
+              </HStack>
+              <Text fontSize="sm">
+                If your use case requires a higher limit, please contact us at{" "}
+                <ChakraLink href="mailto:founders@openpipe.ai">founders@openpipe.ai</ChakraLink>.
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/settings/index.tsx]: I've just created a "support@openpipe.ai" inbox so let's use that one here.

app/prisma/schema.prisma
```diff
@@ -438,6 +439,9 @@ model FineTune {
     projectId String  @db.Uuid
     project   Project @relation(fields: [projectId], references: [id], onDelete: Cascade)
 
+    userId    String?  @db.Uuid
+    user      User? @relation(fields: [userId], references: [id], onDelete: Cascade)
```
corbt: On delete/update this should set to null, not cascade. If someone deletes their account we don't want to delete all the fine-tuned models they trained if there are other users on the same project still.

app/src/utils/rateLimit/concurrencyRateLimits.ts
```diff
@@ -0,0 +1,52 @@
...
+  return requestId?.id;
+};
+
+export const recordOngoingRequestEnd = (id: string | undefined) => {
+  if (id) void kysely.deleteFrom("OngoingRequest").where("id", "=", id).execute();
+};
+
+const checkRateLimitHit = async (projectId: string) => {
+  const result = await kysely
+    .selectFrom("Project as p")
```
corbt, line 30: Does this query actually need to select/join from project at all? Can't it just query the OngoingRequest table directly without joining on Project?

app/src/utils/rateLimit/concurrencyRateLimits.ts
```diff
@@ -0,0 +1,52 @@
...
+      sql<boolean>`CASE WHEN COUNT(req.id) >= p."rateLimit" THEN FALSE ELSE TRUE END`.as(
+        "canProceed",
+      ),
+    )
+    .executeTakeFirst();
+
+  if (!result?.canProceed) {
+    throw new RateLimitError(
+      429,
+      { error: "Completion rate limit exceeded" },
```
corbt: For projects without a credit card:
"Your project has a rate limit of 3 concurrent requests. Add a payment method to automatically increase your rate limit to 20 concurrent requests. [link to project billing page]"

For projects with a credit card:
"Your project has a rate limit of X concurrent requests. Contact us at support@openpipe.ai to increase your rate limit further."

corbt (commented):
Thanks, that was fast work!

Actually let's make this slightly more complex as a way of encouraging folks to add a payment method. For new projects let's set the default rate limit to 3, and then for any project that adds a credit card let's automatically bump the rate limit up to 20.

app/src/utils/rateLimit/concurrencyRateLimits.ts
```diff
@@ -0,0 +1,52 @@
+import { sql } from "kysely";
+import { kysely } from "~/server/db";
+import { v4 as uuidv4 } from "uuid";
+import { RateLimitError } from "openai";
+
+export const recordOngoingRequestStart = async (projectId: string, proceed = true) => {
```
corbt, line 7: Let's add some basic tests for this functionality, similar to the existing tests I have in rateLimits.test.ts

app/src/components/admin/Projects/AdminProjectsTable.tsx
```diff
@@ -150,6 +151,14 @@ const ExtendableArea: React.FC<ExtendableAreaProps> = ({ expandedRow, project })
                 Add credits
               </Button>
             </AddCredits>
+            <Text size="sm">
```
corbt, line 154: Nice, really appreciate you adding this proactively!

app/src/pages/p/[projectSlug]/settings/index.tsx
```diff
@@ -180,6 +181,19 @@ export default function Settings() {
             </VStack>
             <OpenaiApiKeyDisplay />
             <Divider />
+            <VStack alignItems="flex-start">
+              <Subtitle>Rate Limit</Subtitle>
+              <HStack>
+                <Text color="gray.500">Concurrent requests:</Text>
+                <Text fontWeight="bold">{selectedProject?.rateLimit}</Text>
+                <InfoCircle tooltipText="The limits are based by the number of concurrent requests. There is no limit on the total number of requests you can send, but rather on how many you can send at the same time." />
```
Comment (corbt) [inline app/src/pages/p/[projectSlug]/settings/index.tsx]: Instead of putting this in an info box let's just put the text inline similar to the text under "OpenAI API Key".

Text if you haven't added a payment method:
"Your rate limit represents the number of concurrent requests you can send to our servers without getting a 429 "Too Many Requests" error. [Add a payment method](link to add payment method page) to increase your rate limit to 20 concurrent requests automatically."

Text if you have a payment method:
"Your rate limit represents the number of concurrent requests you can send to our servers without getting a 429 "Too Many Requests" error. To increase your rate limit, email [support@openipe.ai](support@openpipe.ai)."

app/src/utils/rateLimit/concurrencyRateLimits.ts
Kovbo: Awesome idea! I must hardcode it by checking if the current rate limit = 3. To check if a project actually has a payment method, we need to call Stripe.

app/src/pages/p/[projectSlug]/settings/index.tsx
Comment (Kovbo) [reply app/src/pages/p/[projectSlug]/settings/index.tsx]: Current projects without a payment method already have a rate limit of 1000. I guess in this place, I also need to hardcode a check if the current rate limit is 3.

app/prisma/schema.prisma
Kovbo: Good idea, changed!

app/src/utils/rateLimit/concurrencyRateLimits.ts
corbt: Ok that seems like a fine workaround short term. If it ends up causing confusion (eg. because we change the default rate limit) we can add a new flag to projects `hasValidPaymentMethod` or something that we update when a payment method is added.

app/src/utils/rateLimit/concurrencyRateLimits.ts
Kovbo: Yes, I think I can just make a subquery on the projects table calculating the number of OngoingRequests.

app/src/utils/rateLimit/concurrencyRateLimits.ts
Kovbo: Addressed all the other comments and tested manually. Will need some time to write tests.

app/src/server/api/routers/payments.router.ts
```diff
@@ -93,6 +94,20 @@ export const paymentsRouter = createTRPCRouter({
           defaultPaymentMethodId = availablePaymentMethodId;
         }
 
+        // Update Rate Limit after a payment method is added
+        if (paymentMethods.data.length) {
```
Kovbo, line 98: It's a weird place to put this logic, but by adding a payment method, we call Stripe directly, not through our API. After that, we hit this endpoint to fetch payment methods and set a default one.

app/src/utils/rateLimit/concurrencyRateLimits.ts
corbt: Ah sorry, my question was whether we needed to reference the project table at all, but of course that was a silly question -- we do need the rate limit field from that table. This query seems reasonable.

app/src/server/api/routers/payments.router.ts
```diff
@@ -93,6 +94,20 @@ export const paymentsRouter = createTRPCRouter({
...
+          const project = await prisma.project.findUniqueOrThrow({
+            where: { id: input.projectId },
+          });
+
+          if (project.rateLimit === 3) {
```
corbt, line 103: Let's define constants for the "3" and "20" and use them everywhere (including on the settings page) so we don't forget to update them somewhere when we change these limits in the future.

corbt (approved):

PR #630, opened 2024-02-23
Tags pg insertion fix

Kovbo:
We’re constantly getting syntax errors when inserting into the “LoggedCallTag” table.

Theory: a user sends incorrect `tags` json.

When a user sends a request with incorrect `tags`, for example, "headers: {"op-tags": '{"1"}'}", we catch this error and return it to a user. But still, we try to insert an empty tags array into a db, getting this error.


corbt (approved):
Ok, let's see if this fixes it!

PR #633, opened 2024-02-26
Rate limit values moved to const

Kovbo:


app/src/utils/rateLimit/const.ts
```diff
@@ -0,0 +1,4 @@
+export const CONCURRENCY_RATE_LIMITS = {
+  BASE_LIMIT: 3,
```
Kovbo, line 2: I cannot export this const directly from [app/src/utils/rateLimit/concurrencyRateLimits.ts](https://github.com/OpenPipe/OpenPipe/pull/633/files#diff-c6bb44bf9fd0f6d915672433e83c390c0706383bf88ad9112d94f1f08ac2fd9b) because that file imports server-side code (db file) that results in an error while importing on the front end. Maybe there is another place where we store such constants?

corbt: Creating a new file to avoid circular deps is probably the best approach here.

corbt (approved):
Thanks!

PR #635, opened 2024-02-26
Upgrade graphile worker version

arcticfly:


corbt (commented):
I think we need to add a `queueName` to the job that counts tokens or something like that?

arcticfly:
> I think we need to add a `queueName` to the job that counts tokens or something like that?

That should be orthogonal (queues do the same thing in both versions).

corbt (approved):

PR #647, opened 2024-02-29
Don't rely on the HF hub at training time

corbt:
They've had a couple of multi-hour downtime incidents in the last month and we don't really need them at training time.


corbt:
Actually this doesn't fully fix the problem since for some reason something inside Axolotl relies on HF as well and breaks when it's down. 🤦

PR #648, opened 2024-02-29
export weights 7 days fix

Kovbo:
Allow export again in 7 days after the file link has expired.


corbt (approved):
That was definitely an oversight on my part, thanks. 🤦

PR #649, opened 2024-03-01
Tool call and content streaming

Kovbo:
It adds support for tool streaming for both providers.

It checks if a user used any `tools` in its input. If yes - a function determines if it is a tool call response or a regular text completion making the necessary deserialization and conversion to OpenAI format.

Test cases include a new `openpipe:test-tool-calls-mixtral-p3` model to test Fireworks.

The code is hard to read, it probably requires some refactoring, but it works.

### DIFF
--- a/app/src/modelProviders/fine-tuned/getAnyscaleCompletion.ts
+++ b/app/src/modelProviders/fine-tuned/getAnyscaleCompletion.ts
@@ -12,7 +12,10 @@ import OpenAI from "openai";
 import { type TypedFineTune } from "~/types/dbColumns.types";
 import { deserializeChatOutput, serializeChatInput } from "./serializers";
 import { env } from "~/env.mjs";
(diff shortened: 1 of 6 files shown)


corbt (approved):
Nice work; I can see why this was hard! Really appreciate you adding the tests, that adds confidence here.

I do suspect/hope there's a way to refactor this to be cleaner and take some of the complexity out of the `getAnyscaleCompletion` and `getFireworksCompletion` functions. But I guess that's hard to do given that `getFireworksCompletion` responds with a different format. I'm ok merging as-is if you can't think of any way to implement this more cleanly.

Kovbo:
@corbt Yes, I did not extract the provider-specific logic outside of the files, and Fireworks logic adds some complexity. I hope they will support OpenAI compatible responses and chat completions API without extra configurations to make it simpler.

PR #652, opened 2024-03-01
remove invoices duplicates

Kovbo:
`pnpm tsx src/server/scripts/removeDuplicateInvoices.ts`
`pnpm tsx src/server/scripts/connectUsageWithInvoices.ts` <- This script may take some time.

### DIFF
--- a/app/src/server/scripts/connectUsageWithInvoices.ts
+++ b/app/src/server/scripts/connectUsageWithInvoices.ts
@@ -1,0 +1,24 @@
+import dayjs from "dayjs";
+import { kysely, prisma } from "../db";
+
+const startOfFeb2024 = dayjs("2024-02-01").startOf("month").toDate();
+const endOfFeb2024 = dayjs("2024-02-01").endOf("month").toDate();
+
+const invoices = await prisma.invoice.findMany({
+  where: {
+    billingPeriod: "Feb 2024",
+  },
+});
+
+for (const invoice of invoices) {
+  await kysely
+    .updateTable("UsageLog")
+    .set({
+      invoiceId: invoice.id,
+    })
+    .where("projectId", "=", invoice.projectId)
+    .where("createdAt", ">=", startOfFeb2024)
+    .where("createdAt", "<=", endOfFeb2024)
+    .where("invoiceId", "is", null)
+    .execute();
+}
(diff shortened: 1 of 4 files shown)


app/src/server/scripts/connectUsageWithInvoices.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const invoices = await prisma.invoice.findMany({
+  where: {
+    billingPeriod: "Feb 2024",
+  },
+});
+
+for (const invoice of invoices) {
+  await kysely
+    .updateTable("UsageLog")
```
corbt, line 15: We should only update usage where it's currently null. We don't want to ever pull usage away from an existing invoice.

app/src/server/scripts/removeDuplicateInvoices.ts
```diff
@@ -0,0 +1,35 @@
+import { prisma } from "../db";
```
corbt, line 1: add a check that we only delete canceled invoices

app/src/server/tasks/worker.ts
```diff
@@ -90,7 +90,7 @@ const runner = await run({
       identifier: chargeInvoices.task.identifier,
       options: {
         backfillPeriod: 1000 * 60 * 60 * 24 * 7,
-        queueName: "process-invoices",
+        queueName: "charge-invoices",
```
corbt, line 93: leave this as the same queue. We don't want to be charging and processing invoices at the same time because could lead to concurrency bugs.

app/src/server/scripts/connectUsageWithInvoices.ts
Kovbo: I think `.where("invoiceId", "is", null)` already does this?

app/src/server/tasks/worker.ts
Kovbo: Oh, I see, sorry!

app/src/server/scripts/removeDuplicateInvoices.ts
Kovbo: Added!

app/src/server/scripts/connectUsageWithInvoices.ts
corbt: Apologies, missed that!

corbt (approved):
Thank you!

PR #657, opened 2024-03-04
[WIP] Local auth for dev environment with credentials provider

rajshirolkar:
In this PR we're adding a Credentials Provider in development environment to have hassle free local auth experience for users running on local.
We accept the email and create a local user with custom session implementation to make it work with our current GitJHub Auth.

image

image

NOTE : 
1. Needs some cleaning up as I have disable some es-lint errors and I'm not particularly proud of it.
2. Needs some thorough testing to detect any session related malfunctions and to check if it does not interfere with existing GitHub Provider session management


corbt:
Nice. I think this might work smoother if it were an "either/or" thing though? So if you have the github auth env variables set, you get the exact same behavior as production, and if you don't then you instead get an email style flow.

app/src/components/nav/AppShell.tsx
```diff
@@ -122,7 +122,7 @@ const NavSidebar = () => {
               as={ChakraLink}
               justifyContent="start"
               onClick={() => {
-                signIn("github").catch(console.error);
```
corbt: How is the behavior in prod affected if we call `signIn` without passing "github" explicitly?

rajshirolkar: The `signIn()` function without arguments defaults to redirecting the user to the NextAuth.js sign-in page, where they can select their preferred sign-in method from the configured options.

Documentation : https://next-auth.js.org/getting-started/client#signin

This is how it would look on prod. Currently on prod we directly go to the GitHub login page.

image

rajshirolkar:
> Nice. I think this might work smoother if it were an "either/or" thing though? So if you have the github auth env variables set, you get the exact same behavior as production, and if you don't then you instead get an email style flow.

Sure. I believe this condition should satisfy it `if (process.env.GITHUB_CLIENT_ID === "your_client_id") `.

app/src/components/nav/AppShell.tsx
corbt: Can this PR be written in such a way that it doesn't affect prod behavior?

rajshirolkar: Sure! I have made the behavior default to "github" in line with current prod setup. Credentials provider will be loaded only when github isnt configured in env.

corbt: I'm a little confused by the updated behavior. Is there any reason to look for the literal "your_client_id", instead of making `GITHUB_CLIENT_ID` an optional field and checking for its presence?

rajshirolkar: Yes! Right now in the  `.env.example` file we ask users to copy,  has a default value for GITHUB environment variables. Hence the literal comparison. However If we're to just check for presence of the field we will have to update the env file as such.

```
# Before
GITHUB_CLIENT_ID="your_client_id"
GITHUB_CLIENT_SECRET="your_secret"

# After
GITHUB_CLIENT_ID=""
GITHUB_CLIENT_SECRET=""
```

I am fine with either of these approaches and I agree with you that checking for presence might be slightly better.

corbt: Let's remove the variables from the example file entirely, mark them as optional in env.mjs, and then just check if they're set or not to choose which branch to take. Thanks!

rajshirolkar: Sure! I have made the changes. Seems a lot cleaner now.

app/src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,4 +1,190 @@
...
 
-export default NextAuth(authOptions);
+const GitHubProvider = ensureDefaultExport(GitHubModule);
+
+export function authOptionsWrapper(req: NextApiRequest, res: NextApiResponse) {
+  const isCredentialsCallback =
+    req.query?.nextauth?.includes("callback") &&
+    req.query.nextauth?.includes("credentials") &&
+    req?.method === "POST";
+  const providers: Provider[] = [
```
Comment (corbt) [inline app/src/pages/api/auth/[...nextauth].ts]: Let's instead construct the providers positively. Eg

```ts
const providers: Provider[] = (env.GITHUB_CLIENT_ID && env.GITHUB_CLIENT_SECRET) ? [ /* github provider */ ] : [ /* local provider */ ]
```

Comment (rajshirolkar) [reply app/src/pages/api/auth/[...nextauth].ts]: Good one. Definitely looks a lot cleaner now!

corbt:
The latest version of this PR still seems to change production behavior.

Steps to reproduce:
1. Have valid Github app credentials set in the .env file
2. Clear cookies on localhost
3. Visit http://localhost:3000

Current behavior on main branch: you see a mostly-empty app shell for a few seconds, then are automatically signed in
Behavior on this branch: you are redirected to a page with a "Sign in with GitHub" link.

Screenshot 2024-03-16 at 10 20 45 PM

rajshirolkar:
Thanks! I have cleaned up the client side variables and from my tests everything is working as expected. I have also updated the Readme and checked-in a `launch.json` file so people can use the debugger for this app.

PR #658, opened 2024-03-04
mixtral function call stream fix

Kovbo:
Removed an unnecessary chunk yield.


corbt (approved):
Thanks, feel free to merge after the small tweak discussed!

PR #662, opened 2024-03-06
Improve NodeEntry loading performance

arcticfly:
Loading the first of a large number of NodeEntry rows for a dataset is a slow proposition right now. Experimentation has revealed that this latency is primarily due to the join operation between NodeEntry and DatasetEntryInput/DatasetEntryOutput rows.

By selecting the ids of filtered NodeEntry rows in one query and performing the join/selection from our input and output tables in another query, we can lose a lot of that latency.

### DIFF
--- a/app/scripts/copy-project.ts
+++ b/app/scripts/copy-project.ts
@@ -123,7 +123,14 @@ await Promise.all([
     "DataChannel",
     `"destinationId" IN (SELECT id FROM "Node" WHERE "projectId" = '${projectId}')`,
   ),
-  copyTable("NodeEntry", `"nodeId" IN (SELECT id FROM "Node" WHERE "projectId" = '${projectId}')`),
+  copyTable(
+    "NodeEntry",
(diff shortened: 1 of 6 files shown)


app/scripts/copy-project.ts
```diff
@@ -123,7 +123,14 @@ await Promise.all([
     "DataChannel",
     `"destinationId" IN (SELECT id FROM "Node" WHERE "projectId" = '${projectId}')`,
   ),
-  copyTable("NodeEntry", `"nodeId" IN (SELECT id FROM "Node" WHERE "projectId" = '${projectId}')`),
```
corbt: Supercedes https://github.com/OpenPipe/OpenPipe/pull/661?

corbt (approved):

app/scripts/copy-project.ts
arcticfly: Forgot to merge that in, thanks

PR #664, opened 2024-03-07
Infrastructure security

Kovbo:
New TLS versions and security headers.
Updated @type/node depending to update a transitive `undici` package from the report.

The most annoying is setting Content Security Headers. It is hard to test if anything is missing. Some packages, for example, `monaco-editor`, call an external CDN under the hood. I manually checked all the pages on staging and did not find any calls so far.

### DIFF
--- a/infra/src/app-service.ts
+++ b/infra/src/app-service.ts
@@ -68,7 +68,10 @@ export const service = new kubernetes.core.v1.Service(
         "service.beta.kubernetes.io/aws-load-balancer-ssl-cert": certificateArn,
         "service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "http",
         "service.beta.kubernetes.io/aws-load-balancer-ssl-ports": "443",
-        "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "60",
+        "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "900",
+        "service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy":
+          "ELBSecurityPolicy-TLS13-1-2-2021-06",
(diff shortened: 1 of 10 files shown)


infra/src/app-service.ts
```diff
@@ -68,7 +68,10 @@ export const service = new kubernetes.core.v1.Service(
         "service.beta.kubernetes.io/aws-load-balancer-ssl-cert": certificateArn,
         "service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "http",
         "service.beta.kubernetes.io/aws-load-balancer-ssl-ports": "443",
-        "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "60",
+        "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "900",
+        "service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy":
+          "ELBSecurityPolicy-TLS13-1-2-2021-06",
+        "service.beta.kubernetes.io/aws-load-balancer-type": "nlb",
```
corbt, line 74: Thanks Bohdan. This specific change worries me slightly -- do you know if there's any downtime associated with the change in load balancer?

Kovbo: Hmm, provisioning a load balancer takes some time, let me test it on staging to measure the downtime.

corbt: if EKS is smart it'll bring up the new LB and service and make sure it's live and then cut over all traffic before tearing down the old one. So there may be no downtime at all. But we should test that theory before deploying in prod. :)

Kovbo: Just tested it, and it seems there is no downtime.

corbt (approved):
Awesome, will get this deployed tonight!

PR #665, opened 2024-03-08
Experimental support for Mistral 7B Instruct v0.2

corbt:
Add experimental support for mistral 7b instruct v0.2, behind a feature flag.


app/scripts/copy-model-weights.ts
```diff
@@ -69,16 +73,39 @@ if (argv.modal) {
...
     `${localPath}/fireworks.json`,
     JSON.stringify({
-      base_model: "accounts/fireworks/models/mixtral-8x7b-instruct-hf",
+      base_model: fwConfig.baseModel,
     }),
   );
 
-  await $$`firectl create model ${model.id} --deploy --wait /tmp/${modelSlug}/`;
+  // const fireworksModelName = model.id;
+  const fireworksModelName = "tmp-delete-4";
```
arcticfly: Do you need to update this?

trainer/src/trainer/upload_to_fireworks.py
```diff
@@ -0,0 +1,143 @@
...
+    with open(adapter_config_path, "r") as f:
+        adapter_config = json.load(f)
+        for key in ["use_rslora", "use_dora"]:  # Check for both keys
+            if key in adapter_config:
+                if adapter_config[key]:
+                    raise Exception(
+                        f"Fireworks does not support the {key} adapter config key"
+                    )
+                else:
+                    del adapter_config[key]
```
arcticfly, line 35: Isn't this a no-op? If I read the code correctly, we throw an exception if this key is present, so deleting it shouldn't make any difference.

arcticfly (approved):

trainer/src/trainer/upload_to_fireworks.py
corbt: Actually, we throw an error if the key is set to True. If it's false, we just delete it from the config (since Fireworks doesn't know what to do with it and throws an error if it's present).

PR #673, opened 2024-03-10
Ensure all eval results are assigned a score

arcticfly:
Fixes a race condition in which some eval results copied their score from an incomplete result.

We're also now substituting in model titles on the fly when loading head to head comparisons, so changed model names will be reflected in all associated comparison explanations.


app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -140,14 +165,15 @@ export const evaluateTestSetEntries = defineTask<EvalKey>({
...
-          .on("der.status", "=", "COMPLETE"),
+          .on("der.status", "=", "COMPLETE")
+          .on("der.score", "!=", null),
       )
       .innerJoin("DatasetEvalResult as comparisonDer", (join) =>
         join
           .onRef("comparisonDer.nodeEntryInputHash", "=", "ne.inputHash")
-          .on("comparisonDer.nodeEntryOutputHash", "=", nodeEntry.outputHash)
           .on("comparisonDer.datasetEvalOutputSourceId", "=", secondOutputSourceId)
-          .on("der.status", "=", "COMPLETE"),
```
arcticfly, line 150: This line was the primary cause of the bug. Instead of checking if `comparisonDer` was complete, we were only checking to see if `der` was complete, which might have been a DatasetEvalResult that was not compared against`comparisonDer`. This meant that in many cases, `comparisonDer` was not actually complete, so the score we copied from it was null.

app/src/server/tasks/evaluateTestSetEntries.task.ts
```diff
@@ -103,11 +125,14 @@ export const evaluateTestSetEntries = defineTask<EvalKey>({
 
     const firstResultHandled =
       (firstResult.status === "IN_PROGRESS" || firstResult.status === "COMPLETE") &&
+      // backfill results if they were created without a score (3/9/2024 bugfix)
+      firstResult.score !== null &&
```
arcticfly, line 129: Any eval with null comparison scores as a result of this bug should automatically fix itself the next time that eval is run after this change is merged in.

corbt (approved):

PR #680, opened 2024-03-11
Database replication and encryption

Kovbo:
Logic:
1. Create an empty encrypted database _- Executes automatically_
2. Copy all the data from the master database using Amazon DMS.  _- Executes automatically_
   2.1 Create required DMS Roles and Policies, Subnet Group.
   2.2 Create DMS Replication Instance, Source and Target endpoints. 
   2.3 Create DMS Migration Task.
3. Enable `logical replication` by applying the `dbParameterGroup` to the master DB _- Executes automatically_
4. Restart master DB - **Executes manually**
5. Start migration task - **Executes manually**
6. Add indexes to the encrypted copy - **Executes manually**
7. Add indexes and foreign keys (if necessary) to the encrypted DB copy  - **Executes manually**
8. Test, switch db, and cleanup - will be done later in the next pull request.

 It will copy all the tables from the `public` schema. Note: it cannot copy the `graphile_worker` schema.


infra/src/database.ts
```diff
@@ -46,6 +57,205 @@ const dbInstance = new aws.rds.Instance(nm("app"), {
   finalSnapshotIdentifier: nm("app-db-final-snapshot"),
 });
 
+const encryptedDbInstance = new aws.rds.Instance("app-encrypted", {
```
corbt: use the `nm` helpers so we see the environment name in the resource names in AWS.

infra/src/database.ts
```diff
@@ -46,6 +57,205 @@ const dbInstance = new aws.rds.Instance(nm("app"), {
...
+  password: password.result,
+  dbName,
+  instanceClass: "db.m7g.2xlarge",
+  allocatedStorage: 1210,
+  maxAllocatedStorage: 2000,
+  storageType: "io1",
+  iops: 12000,
+  engine: "postgres",
+  engineVersion: "15.5",
+  parameterGroupName: "default.postgres15",
```
corbt: As long as we're making big db changes we should upgrade to pg 16 at the same time.

infra/src/database.ts
```diff
@@ -46,6 +57,205 @@ const dbInstance = new aws.rds.Instance(nm("app"), {
...
+  publiclyAccessible: true,
+  applyImmediately: true,
+  backupRetentionPeriod: 30,
+  performanceInsightsEnabled: true,
+  multiAz: isProd,
+  skipFinalSnapshot: !isProd,
+  // snapshotIdentifier: isProd ? undefined : "app-pl-stage-test-snapshot",
+  finalSnapshotIdentifier: nm("app-db-encrypted-final-snapshot"),
+  storageEncrypted: true,
+  identifier: "app-encrypted-db-instance",
```
corbt: We should just let pulumi set the name. Hardcoding it like this will fail in prod anyway, since it'll try to use the same name for the existing staging one and fail because the name has to be unique.

infra/src/database.ts
```diff
@@ -46,6 +57,205 @@ const dbInstance = new aws.rds.Instance(nm("app"), {
...
+      principals: [
+        {
+          identifiers: ["dms.amazonaws.com"],
+          type: "Service",
+        },
+      ],
+    },
+  ],
+});
+const dms_access_for_endpoint = new aws.iam.Role("dms-access-for-endpoint", {
```
corbt: Always use the `nm` helper for resource names unless there's a reason not to.

infra/src/database.ts
Kovbo: Upgraded and addressed all other changes!

corbt (approved):
Awesome, looks good!

PR #686, opened 2024-03-13
Sentry projectId and key added

Kovbo:
after

### DIFF
--- a/app/src/server/api/external/openApiTrpc.ts
+++ b/app/src/server/api/external/openApiTrpc.ts
@@ -89,7 +89,15 @@ const sentryMiddleware = t.middleware(async ({ ctx, next }) => {
     attachRpcInput: true,
   });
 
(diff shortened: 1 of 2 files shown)


corbt (changes requested):
Specifically, I'm hoping we can set the user to the project owner. That way in eg. the view below we can quickly see the count of how many users (project owners) are affected by each issue.
 
Screenshot 2024-03-12 at 7 33 12 PM

Kovbo:
> Specifically, I'm hoping we can set the user to the project owner. That way in eg. the view below we can quickly see the count of how many users (project owners) are affected by each issue.

Nice idea, added!
after
I think it requires an additional db query per request. If it adds a delay, we could use a projectId as a user identifier.

corbt (approved):
Great, thanks!

PR #694, opened 2024-03-16
[WIP] Evaluation Model options (+external models)

Kovbo:
-Adds the ability to choose an evaluation model.
-Fixes price calculation
-Gives an ability to add external models (work in progress)
after


app/src/server/api/routers/externalModel.router.ts
```diff
@@ -0,0 +1,40 @@
...
+  defaultEvaluationModel,
+  findPredefinedEvaluationModel,
+} from "~/utils/externalModels/evaluationModels";
+
+export const externalModelRouter = createTRPCRouter({
+  get: protectedProcedure
+    .input(
+      z.object({
+        id: z.string().optional(),
+        projectId: z.string().optional(),
```
corbt, line 15: Any reason not to make `projectId` required?

corbt:
This looks pretty good to me! I'd like @arcticfly to review as well since he wrote most of the code you're touching here and is likely to think of things that I haven't.

app/src/components/evals/EvalContentTabs/Settings/Settings.tsx
```diff
@@ -117,39 +119,26 @@ const Settings = () => {
...
+      usedEvaluationModel && setSelectedEvaluationModel(usedEvaluationModel);
     }
   }, [datasetEval, setName, setInstructions, setNumRows, setIncludedModelIds]);
 
-  useEffect(() => reset(), [datasetEval, reset]);
+  useEffect(() => {
+    usedEvaluationModel && setSelectedEvaluationModel(usedEvaluationModel);
+  }, [usedEvaluationModel]);
+
+  useEffect(() => void reset(), [datasetEval, reset]);
```
arcticfly, line 141: Is reset async? I don’t think you need to use void when calling it.

app/src/components/evals/EvalForm/EvalForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+    numComparisons: number;
+  };
+  evaluationModelState: {
+    selectedEvaluationModel: ExternalModel;
+    setSelectedEvaluationModel: React.Dispatch<React.SetStateAction<ExternalModel>>;
+  };
+  instructionsState: {
+    instructions: string;
+    setInstructions: React.Dispatch<React.SetStateAction<string>>;
+  };
```
arcticfly, line 35: It doesn’t really hurt to add these explicit typings, but is there any reason not to just use `(instructions: string) => void`?

app/src/components/evals/EvalForm/EvalForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+}
+
+const EvalForm: React.FC<EvalNameProps> = ({
+  datasetId,
+  nameState,
```
arcticfly, line 40: Nit: I would probably deconstruct the state props right here in the function signature. I think you’ll have fewer lines of code overall, and I don’t think you gain much from doing it down in the function body.

app/src/components/evals/EvalForm/EvalForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+  evaluationModelState,
+  instructionsState,
+}) => {
+  const { name, setName } = nameState;
+  const { includedModelIds, setIncludedModelIds } = modelState;
+  const { numRows, setNumRows, numComparisons } = rowsState;
+  const { selectedEvaluationModel, setSelectedEvaluationModel } = evaluationModelState;
+  const { instructions, setInstructions } = instructionsState;
+
+  const additionalInputPayloadTokens = 40;
```
arcticfly, line 52: Good idea to account for additional tokens. How did you calculate this number?

app/prisma/schema.prisma
```diff
@@ -864,3 +865,18 @@ model CachedProcessedEntry {
...
     @@index([projectId])
 }
+
+model ExternalModel {
+    id String @id @default(uuid()) @db.Uuid
+
+    name String
+    displayName String
+    endpoint  String
+    apiKey  String
```
arcticfly, line 875: For things like calculating the price of evals, won’t we need to know price per input and output token?

app/src/components/evals/EvalForm/EvalForm.tsx
arcticfly: Also, I would uppercase these const variables and declare them above the function to increase readability and communicate that they aren’t derived from the inputs to this function.

app/src/server/api/routers/externalModel.router.ts
arcticfly: Same question for `id`

app/src/server/api/routers/externalModel.router.ts
```diff
@@ -0,0 +1,40 @@
...
+      const { id, projectId } = input;
+      if (projectId) {
+        await requireCanModifyProject(projectId, ctx);
+      }
+
+      return getEvaluationModel(id, projectId);
+    }),
+});
+
+export const getEvaluationModel = async (id?: string, projectId?: string) => {
```
arcticfly, line 29: When will we call this function with an undefined id?

app/src/server/utils/nodes/startTestJobs.ts
```diff
@@ -113,6 +115,7 @@ export const startTestJobsForEval = async ({
         const secondOutputSourceId = datasetEval.outputSources[j]?.id as string;
         evalsToRun.push({
           nodeEntryId: datasetEvalNodeEntry.nodeEntryId,
+          evaluationModelId: evaluationModelId,
```
arcticfly, line 118: Nit: this could be replaced with just `evaluationModelId,`

app/src/utils/externalModels/allModels.ts
```diff
@@ -0,0 +1,57 @@
+import { ExternalModel } from "@prisma/client";
+
+export const predefinedModelNames = {
+  GPT_3_5_TURBO_1106: "gpt-3.5-turbo-1106",
+  GPT_4_0613: "gpt-4-0613",
+  GPT_4_1106_PREVIEW: "gpt-4-1106-preview",
+  GPT_4_0125_PREVIEW: "gpt-4-0125-preview",
+};
+
+// export type externalModel = {
```
arcticfly, line 10: Remove

app/src/utils/externalModels/allModels.ts
```diff
@@ -0,0 +1,57 @@
+import { ExternalModel } from "@prisma/client";
+
+export const predefinedModelNames = {
+  GPT_3_5_TURBO_1106: "gpt-3.5-turbo-1106",
```
arcticfly, line 4: Nit: I would make this an enum

app/src/utils/externalModels/evaluationModels.ts
```diff
@@ -0,0 +1,19 @@
...
+  predefinedModelNames.GPT_4_0613,
+  predefinedModelNames.GPT_4_1106_PREVIEW,
+  predefinedModelNames.GPT_4_0125_PREVIEW,
+];
+
+export const predefinedEvaluationModels: ExternalModel[] = predefinedExternalModels.filter(
+  (model) => evaluationModels.includes(model.name),
+);
+
+export const defaultEvaluationModel = predefinedEvaluationModels.filter(
```
arcticfly, line 14: Use `.find()` here as well.

app/src/utils/externalModels/evaluationModels.ts
```diff
@@ -0,0 +1,19 @@
+import { ExternalModel } from "@prisma/client";
+import { predefinedModelNames, predefinedExternalModels } from "./allModels";
+
+const evaluationModels = [
```
arcticfly, line 4: Why are these different from predefined models?

arcticfly (commented):
Good work refactoring the EvalForm into its own component. Could you include a screenshot of an eval result from GPT-4 0613 when the entries to evaluate were too long for its context?

app/src/server/api/routers/externalModel.router.ts
Kovbo: Instead of having two different endpoints, one of which gets a default predefined model and another gets a model added by the user, I use a single endpoint with optional parameters. Calling a function without parameters returns a default model (GPT-4-Turbo).

app/prisma/schema.prisma
Kovbo: Yes, I don't think we can't get the price from the endpoint/model name. We could ask the client to add this information as an option, but we cannot keep it up to date. 
The current logic does not show the price for external models added by a user.
Do you think we should add an option to a user to enter the price per token?

app/src/server/api/routers/externalModel.router.ts
arcticfly: I understand that this endpoint is going to be used to return both predefined models and models stored in the database, and that makes sense. Under what circumstances is it better to call this endpoint without any parameters, instead of just referencing the default predefined model?

app/prisma/schema.prisma
arcticfly: I know that some other companies ask the user to input token prices, and it seems to work well for them.

app/src/server/api/routers/externalModel.router.ts
Kovbo: You are right, I will probably change this API while working on the actual implementation of custom external models (which I plan to do tomorrow).
The current idea was to put the logic of getting an instance of a model into a single function that determines which model to return instead of adding additional logic to every place where we get a model.
But yeah, it does not make sense when we have different default models for different places.

app/src/utils/externalModels/evaluationModels.ts
Kovbo: `evaluationModels` is a subset of all `predefinedExternalModels`.
For example, we do not use `gpt-3.5-turbo` for evaluations, but we will use it in other places.
Did I understand the question correctly?

app/prisma/schema.prisma
Kovbo: Sure, I will add it then! It might be useful in certain scenarios.

app/src/components/evals/EvalForm/EvalForm.tsx
Kovbo: Kyle precalculated these numbers approximately.
I think I can also calculate the `additionalInputPayloadTokens` based on the `instructions` input. It may be relevant if clients give large instructions for evaluation.

Uppercase is a nice idea, thank you!

app/src/components/evals/EvalContentTabs/Settings/Settings.tsx
Kovbo: Oh yes, I forgot to remove it, thank you!

app/src/components/evals/EvalForm/EvalForm.tsx
Kovbo: Yes, nice idea, I like it!

app/src/components/evals/EvalForm/EvalForm.tsx
Kovbo: Let me try!

app/src/server/utils/nodes/startTestJobs.ts
Kovbo: Thanks! Will do!

app/src/utils/externalModels/evaluationModels.ts
arcticfly: Yes, you answered correctly. However, I don't see any reason not to allow 3.5 as an evaluation model. It would be good at doing certain types of evaluations, and would cost much less. @corbt do you see any reason to omit it?

Kovbo: I think we were initially discussing supporting GPT 4 only for evaluations. But I agree; I do not see any downsides to adding GPT 3.5.

arcticfly: As an aside, I would either create the `evaluationModels` array inside `predefinedEvaluationModels` instead of storing it as an external variable, or change the name to be `evaluationModelNames`, since models means something else in this context.

PR #697, opened 2024-04-01
Migrated dataset upload infra from azure to s3

rajshirolkar:
Dataset uploads will now use AWS S3 bucket instead of Azure Blob storage.


app/src/utils/aws/website.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Get a presigned URL for this part
+    const { presignedUploadUrl } = await api.client.datasets.getPresignedUploadUrlForPart.query({
+      projectId,
+      filename: blobName,
+      partNumber,
+      uploadId: uploadId as string,
+    });
+
+    // Step 3: Upload the part
+    const uploadPartPromise = await fetch(presignedUploadUrl, {
```
corbt, line 42: I'm a bit confused here. You `await`, this, but then below also have an `await Promise.all(partUploadPromises)`. Is one of those `await`s unnecessary?

rajshirolkar: I was facing issues with the ETag and parts values while fetching. Removed the await on fetch so the array is full of promises now. Also removed the unused `completeMultipartUploadResponse` variable. Thanks for catching! Will clean up rest of the code too.

